PR #19, opened 2019-01-31
action playback

amandlek:
- Added functionality for deterministic open loop playback of actions from demonstrations. Note that this only applies to new demonstrations that are collected with these changes (old demonstrations will not playback correctly).
- Fixed a bug where we were possible dropping a few intermediate (s, a) pairs during data collection.

### DIFF
--- a/robosuite/scripts/playback_demonstrations_from_hdf5.py
+++ b/robosuite/scripts/playback_demonstrations_from_hdf5.py
@@ -30,7 +30,7 @@ if __name__ == "__main__":
         ),
     )
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


robosuite/scripts/playback_demonstrations_from_hdf5.py
```diff
@@ -23,6 +29,10 @@
             robosuite.models.assets_root, "demonstrations/SawyerNutAssembly"
         ),
     )
+    parser.add_argument(
+        "--use_actions",
```
jirenz: Ideally you want "--use-actions".

PR #32, opened 2019-06-07
fix reset issue for sawyer and baxter with grippers

youngwoon:
This fixes an issue in reset function.
In the original code, `self._ref_joint_gripper_right_actuator_indexes` and `self._ref_joint_gripper_left_actuator_indexes` contain indexes of the left wrist.
`self._ref_gripper_right_joint_pos_indexes` and `self._ref_gripper_left_joint_pos_indexes` should be used instead of them.


yukezhu:
Thank you @youngwoon !

PR #33, opened 2019-06-18
fixing action replay from demos

amandlek:
- some small fixes to fix deterministic action replay from collected demonstrations

### DIFF
--- a/robosuite/wrappers/ik_wrapper.py
+++ b/robosuite/wrappers/ik_wrapper.py
@@ -109,8 +109,13 @@ class IKWrapper(Wrapper):
                 velocities = self.controller.get_control()
                 if self.env.mujoco_robot.name == "sawyer":
                     low_action = np.concatenate([velocities, action[7:]])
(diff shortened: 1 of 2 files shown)


robosuite/wrappers/ik_wrapper.py
```diff
@@ -105,11 +105,12 @@ def step(self, action):
...
-            velocities = self.controller.get_control()
-            if self.env.mujoco_robot.name == "sawyer":
-                low_action = np.concatenate([velocities, action[7:]])
-            else:
-                low_action = np.concatenate([velocities, action[14:]])
+            if i + 1 < self.action_repeat:
+                velocities = self.controller.get_control()
+                if self.env.mujoco_robot.name == "sawyer":
+                    low_action = np.concatenate([velocities, action[7:]])
+                else:
```
yukezhu: change this else condition to an elif condition to check for name "baxter". throw an exception if the name if unrecognized.

PR #47, opened 2019-10-26
Merge extra envs

cremebrule:
Added the following envs:
Sawyer:
`SawyerFreeSpaceTraj`,`SawyerDoor`,`SawyerWipeForce`, `SawyerWipePegs`, `SawyerWipeTactile`,`SawyerWipe3DTactile`

Panda:
`PandaFreeSpaceTraj`,`PandaDoor`,`PandaWipeForce`, `PandaWipePegs`, `PandaWipeTactile`,`PandaWipe3DTactile`

Added two "base" envs from which the above user-facing envs extend:
`SawyerRobotArmEnv`, `PandaRobotArmEnv`

Added multipurpose controller `arm_controller.py` which includes osc functionality. Note: Only the above new envs rely on this controller file; a future merge request will integrate this controller functionality to the remaining environments.

* It is useful to note that while these new envs have greater (controller) functionality than the current pre-existing robosuite envs, they have been integrated in such a way that maintains backwards-compatibility(i.e.: no change to required arguments when calling `robosuite.make`). While a more complex controller (e.g.: position/orientation) may be specified during the `robosuite.make` call, it defaults to a `joint_velocity` controller which is in accordance with prior envs. However, small discrepancy exists -- since the new controller does not include the gripper as part of its action space, the expected `action` fed into `env.step(action)` should be equal to the dof of the robot currently being controlled (as opposed to dof+1 as in previous envs).

Added default config files holding parameters relevant to specific tasks and the various controllers contained in `arm_controller.py`. Included a README that specifies the functionality of each config parameter.

A few TODOs remaining as documentation questions and to clarify desired protocol (see TODOs for specifics).


yukezhu:
There seems to be a lot of redundant and unnecessary files in this PR. I would suggest a major refactoring and cleanup before code review.

PR #58, opened 2020-02-16
controller and task paths are absolute now

amandlek:
- Made it easier to specify new controller and task configs by passing absolute path rather than relative.


roberto-martinmartin:
Please, change the parameter description accordingly

PR #64, opened 2020-02-28
adding script to tune camera

amandlek:
- Wrote a script to easily tune cameras by using the keyboard in an interactive viewer session.

### DIFF
--- a/robosuite/scripts/tune_camera.py
+++ b/robosuite/scripts/tune_camera.py
@@ -1,0 +1,259 @@
+"""
+Convenience script to tune a camera view in a mujoco environment.
+Allows keyboard presses to move a camera around in the viewer, and
+then prints the final position and quaternion you should set
+for your camera in the mujoco XML file.
+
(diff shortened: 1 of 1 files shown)


robosuite/scripts/tune_camera.py
```diff
@@ -0,0 +1,259 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--env",
+        type=str,
+        default="SawyerLift",
+    )
+    args = parser.parse_args()
+
+    # read camera XML tag from user input
+    inp = input("Please paste the current camera xml tag below (e.g. <camera ... />):\n")
```
yukezhu: This script is very functional but not well documented. Consider the following two options (or both):

1. Provide a concrete example, e.g., go to a MJCF file in /models/assets/arenas and copy a camera tag
2. Allow blank input, which uses a default camera initial pose (e.g., the frontview)

Issue #35, opened 2019-07-02
Potential issue with UniformRandomSampler


eugval:
Hello,

That repository is fantastic, thank you for the good work!

I have noticed in the UniformRandomSampler under models.tasks.placementsampler that the z_rotation argument default is `"random"`, in the SaywerLift environment it is passed as `True` , and in the `sample_quat` function where it is used it seems that the only supported options are `[None, Iterable,  scalar]`.

In case I am misunderstanding something, and because I am not sure what the intended defaults were originally,   I am just raising an issue here for now!

![image](https://user-images.githubusercontent.com/24697174/60501897-b8346a80-9cb4-11e9-8c6b-7a8ef22bb4be.png)

![image](https://user-images.githubusercontent.com/24697174/60501950-cbdfd100-9cb4-11e9-8803-607c91d872e6.png)

Thanks!

kaixindelele:
> Hello,
> 
> That repository is fantastic, thank you for the good work!
> 
> I have noticed in the UniformRandomSampler under models.tasks.placementsampler that the z_rotation argument default is `"random"`, in the SaywerLift environment it is passed as `True` , and in the `sample_quat` function where it is used it seems that the only supported options are `[None, Iterable, scalar]`.
> 
> In case I am misunderstanding something, and because I am not sure what the intended defaults were originally, I am just raising an issue here for now!
> 
> ![image](https://user-images.githubusercontent.com/24697174/60501897-b8346a80-9cb4-11e9-8c6b-7a8ef22bb4be.png)
> 
> ![image](https://user-images.githubusercontent.com/24697174/60501950-cbdfd100-9cb4-11e9-8803-607c91d872e6.png)
> 
> Thanks!

when I debug this value,I found the quat is the same every time.

amandlek:
Thanks for letting us know! You are both correct - this is a bug. Currently, passing the True value makes the sampler interpret the input as a fixed z-rotation of 1 radian. We will fix this soon, but the simple work around if you want uniformly sampled rotations is to just pass a value of None for that parameter.

PR #65, opened 2020-03-03
Fix lift initialization

amandlek:
- Fixing a bug where cube for lift task is always initialized with constant rotation.


yukezhu:
Close #35 #65

PR #68, opened 2020-03-28
Interactive objects

danfeiX:


robosuite/environments/sawyer_lift.py
```diff
@@ -727,31 +723,51 @@ class SawyerLiftPositionTarget(SawyerLiftPosition):
...
         self._goal_radius_low = goal_radius_low
         self._goal_radius_high = goal_radius_high
         self._goal_grid = None
-        if kwargs.get('eval_mode', False) or hide_target:
-            self._target_rgba = (0., 0., 0., 0.,)
+        self._hide_target = hide_target
+
+        self._target_name = 'cube_target'
+        self._object_name = 'cube'
+        self.interactive_objects = {}
```
amandlek: what's the point of initializing with an empty dict here?

robosuite/environments/sawyer_lift.py
```diff
@@ -826,10 +844,41 @@ def _reset_internal(self):
...
-        cube_pos[1] += radius * np.sin(angle)
-        print(cube_pos)
-        self._set_target(pos=cube_pos)
+        target_pos = cube_pos.copy()
+        target_pos[0] += radius * np.cos(angle)
+        target_pos[1] += radius * np.sin(angle)
+        self._set_target(pos=target_pos)
+        for _, o in self.interactive_objects.items():
+            o.reset()
+        # with open('default.xml') as f:
```
amandlek: still need these comments, or should they be deleted?

robosuite/models/objects/interactive_objects.py
```diff
@@ -0,0 +1,255 @@
...
+        for cond, func in self._on_enter_state_funcs:
+            if satisfies(self.state, cond) and not satisfies(prev_state, cond):
+                func()
+
+        for cond, func in self._on_exit_state_funcs:
+            if not satisfies(self.state, cond) and satisfies(prev_state, cond):
+                func()
+
+
+class StatefulInteractiveObject(StatefulObject, InteractiveObject):
```
amandlek: could you explain the need for multiple inheritance here?

robosuite/models/objects/interactive_objects.py
```diff
@@ -0,0 +1,255 @@
...
+        return self.sim.model.body_id2name[self._body_id]
+
+    @property
+    def body_geom_ids(self):
+        return EU.bodyid2geomids(self.sim, self._body_id)
+
+    def reset(self):
+        self._last_event_step = 0
+
+    def _on_contact(self, contacting_body_ids):
```
amandlek: having trouble with understanding what exactly the difference is between InteractiveObject and StatefulObject. For example, what is the difference between the on_contact here vs. enter and leaving state? Are you factorizing functionality into doing things when enter / leave contact (i.e. setting state) vs. enter / leave state (i.e. state change)?

robosuite/environments/sawyer_lift.py
danfeiX: i'll clean these up

robosuite/environments/sawyer_lift.py
danfeiX: cuz interactive_objects is a dict?

robosuite/models/objects/interactive_objects.py
danfeiX: It doesn't have to be. Just thought this makes more sense because StatefulObject and InteractiveObject serve different purposes.

robosuite/models/objects/interactive_objects.py
danfeiX: states and events that cause states to change are different things. StatefulObject maintains object state and calls function when the state changes, but doesn't decide what event would change the state. InteractiveObject defines basic callback functions, and the sub-classes would determine how each event would change the state.

robosuite/environments/sawyer_lift.py
amandlek: I meant that the whole dict can be defined where you're adding the keys

PR #72, opened 2020-04-02
Refactor arm

cremebrule:
This PR builds upon the `merge_controller_v2` PR.

Major updates to robosuite structure:

- Support for multi-arm environments
- Modularization of environments
   - Robots and tasks are now decoupled
- Minimized redundant code; new environment tasks now can be created with much less overhead

Specifically, the following structure is now applied:

- A `robosuite.make(...)` call returns an environment object. All environments inherit from the base `RobotEnv` class which, in turn, inherits from `MujocoEnv` class.
- During the `make` call, the following parameters MUST be specified:
   - `env_name`: This is the same as before in the old robosuite API (e.g.: `PandaLift`, `SawyerNutAssembly`, etc.). Now, however, the `env_name` is decoupled from the robot(s), so that available options are robot-agnostic (e.g.: `Lift`, `TwoArmLift`, `NutAssembly`, etc.)
   - `robots`: This specifies which robots to instantiate within the environment. Available options are `Panda`, `Sawyer`, or `Baxter`. If the specified `env_name` supports multiple arms, then the appropriate number of arms must be specified via a list (e.g.: for `TwoArmLift`, a list of single armed robots like `['Panda', 'Panda']` or a single bimanual robot like `Baxter` is acceptable).
   - The other arguments from the old robosuite API have been preserved; however, in the case of multi-armed environments, there is the option to set a single setting (per argument) across all arms, or individually set for each arm. For example, the argument `has_offscreen_renderer` is now `has_offscreen_renderers`, and can either be specified with a single value (that will be broadcast across all arms within the environment), or specified per-arm via a list.
   - The following options only apply to multi-armed environments:
      - `env_configuration`: If specified, sets the configuration of the arms setup. Choices are `bimanual` (only applicable for single bimanual robot), `single-arm-parallel`, which places two single-armed robots side by side, and `single-arm-opposed` (default), which places two single-armed robots facing towards each other
   - For detailed information about all environment arguments, please see the argument docstring in the appropriate environment.
- During creation of environment `env`, the environment class segments the input arguments and instantiates individual robots (which all inherit from the `Robot` base class), which are stored in the environment-level attribute `env.robots` array. It is within the `Robot`-inherited classes that controllers are instantiated and mujoco models are loaded. Similarly, during an `env.step(action)` call, the action is parsed into its appropriate components to distribute to the correct robot.
- Many of the redundant attributes, methods, and properties in the old robosuite environments have been redefined within the `Robot`-based classes. In this way, much of the boilerplate overhead has been automated and stripped away from the environment level, accelerating environment development and creation.


yukezhu:
So I was looking at the the design of `options["controller_configs"]`. I wonder if it should be a config of each _robot_, instead of a config of each _environment_ (our current design)? This caused problems when I was trying to use a joint-space controller for Sawyer and OSC for Panda in a TwoArmLift task.

cremebrule:
Hmmm I thought I already included that feature? All of the options should let you either input a single value (which will be broadcast across all robots in the environment) or a list of specific per-robot configurations for each option. So if you wanted your proposed configuration of Joint Space for Sawyer and OSC for Panda, you’d set e.g. controller_configs = [config_sawyer, config_panda]

If you’re referencing the Demo Device script specifically, then yeah that does need to be updated to allow for per-robot configurations at the command-line argument level

yukezhu:
I see. I didn't realize that we can make `controller_configs` a list as well. One issue I found is that `ndim` is still set to 7 when I initiate an EE_OSC_POSITION controller. Anything wrong here?

image

cremebrule:
Nope, “ndim” refers to the number of arm joints the robot has — it’s used by controller_factory to instantiate an interpolator if requested, but it’s not used at all by the actual controller. The control_dim attribute that’s part of all the controller classes is probably what you’re looking for (though I just checked and realized that IK doesn’t have that attribute implemented).

yukezhu:
Given so many variables related to DoFs and action dimensions, it'd be better to use a more discriminative name.

cremebrule:
Yes, agreed. I added this (“disambiguating naming scheme, e.g.: qpos / qvel/ joint_indexes / etc.”) to the improvements doc.

robosuite/controllers/ik.py
```diff
@@ -226,9 +226,11 @@ def setup_inverse_kinematics(self, load_urdf=True):
         # get paths to urdfs
         self.robot_urdf = pjoin(
             os.path.join(robosuite.models.assets_root, "bullet_data"),
-            "{}_description/urdf/{}_arm.urdf".format(self.robot_name, self.robot_name)
+            "{}_description/urdf/{}_arm.urdf".format(self.robot_name.lower(), self.robot_name.lower())
         )
 
+        print(self.robot_name.lower())
```
yukezhu, line 232: is this line needed?

PR #80, opened 2020-04-11
Models fix 2

roberto-martinmartin:


yukezhu:
image
image
Some collision meshes are too small, leading to collisions of visual meshes (not a serious issue, but looks a bit weird)

yukezhu:
The PR looks good to me. I spotted the following two issues:

1. Some of the collision meshes are too small, causing obvious penetrations of the visual meshes. I was unsure if there's an easy, automatic fix instead of going to each MJCF and edit the collision meshed manually. This is a relatively lower priority. We can add this to our `Enhancement` feature lists.

2. I am in favor of removing the `PushingRethinkGripper` for the following three reasons:
- It's essentially the RethinkGripper of fixed "closed" state. Do we have to make a new class for this? At least, it should be called something like `ClosedRethinkGripper`, which is independent of any task.
- The implementation of the `PushingRethinkGripper` class is incorrect. The gripper should have no action and zero dof.
- It is not used by any task in `robosuite`.

What do you think? @roberto-martinmartin

roberto-martinmartin:
> The PR looks good to me. I spotted the following two issues:
> 
> 1. Some of the collision meshes are too small, causing obvious penetrations of the visual meshes. I was unsure if there's an easy, automatic fix instead of going to each MJCF and edit the collision meshed manually. This is a relatively lower priority. We can add this to our `Enhancement` feature lists.
> 2. I am in favor of removing the `PushingRethinkGripper` for the following three reasons:
> 
> * It's essentially the RethinkGripper of fixed "closed" state. Do we have to make a new class for this? At least, it should be called something like `ClosedRethinkGripper`, which is independent of any task.
> * The implementation of the `PushingRethinkGripper` class is incorrect. The gripper should have no action and zero dof.
> * It is not used by any task in `robosuite`.
> 
> What do you think? @roberto-martinmartin

About the collision meshes: they are taken from the URDF of the sawyer provided by rethink. however, this urdf is not optimized for simulation (ROS/Gazebo is not the best for that). An easy alternative is to do what panda has: low poly convex meshes for collisions. I could give it a try. If it is just transforming the original ones, it should be easy to do. Maybe only for the last 3-4 links + gripper (we are not going to do manipulation with the first 3 links)

About the pushing gripper, I'm in favor of removing it. I'm not using it. @amandlek ?

amandlek:
Not using it either - we should remove it.

yukezhu:
Sounds good. I will remove it and merge this PR tonight.

@roberto-martinmartin <https://github.com/roberto-martinmartin>, we can
revisit the idea of making low-poly collision meshes later. I did this with
the mesh simplification tool in Blender before, it's easy and fast. I will
add this as a v1.0 feature in the doc.

PR #83, opened 2020-04-17
Fixing OSC mass matrix indexing

amandlek:
- The current indexing of the mass matrix is wrong. This should fix it.


yukezhu:
I don't quite understand the difference between `qpos_index`, `qpos_vel`, and `joint_index`. In what scenarios they are different? In my experiments they had the same values.

amandlek:
Try using a position indicator - that will change the qpos and qvel index values.

yukezhu:
Do you mean that position indicators have `qpos` but not `qvel`? I found using `qvel` is confusing here. Why not `joint_index`?

amandlek:
The qvel indices are how you are supposed to index the mass matrix, as far as I understand. Perhaps @cremebrule @roberto-martinmartin could comment.

yukezhu:
So are we ready to merge this fix?

cremebrule:
Looks good to me.

PR #90, opened 2020-05-14
[WIP] Domain Randomization

amandlek:
- Basic Domain Randomization functionality.
    - This includes texture, lighting, material, and camera randomization.
- Add "name" property when constructing generated objects.
- Add textures to robots and grippers for texture randomization.
- Relevant files to look at:
    - scripts/demo_domain_randomization.py
    - wrappers/domain_randomization_wrapper.py
    - utils/mjmod.py
-TODOs
    - Decide on DR wrapper options and interface.
        - e.g. local perturbations, types of modification for texture, camera, etc.
    - Implement local perturbation support instead of randomizing uniformly.
    - Decide on if we want textures added to objects by default (except for visual objects) and if we want randomization to happen on object groups (for objects with more than one geom).

### DIFF
--- a/robosuite/models/objects/generated_objects.py
+++ b/robosuite/models/objects/generated_objects.py
@@ -12,6 +12,7 @@ class PotWithHandlesObject(MujocoGeneratedObject):
 
     def __init__(
         self,
+        name,
         body_half_size=None,
         handle_radius=0.01,
         handle_length=0.09,
@@ -22,7 +23,7 @@ class PotWithHandlesObject(MujocoGeneratedObject):
         solid_handle=False,
         thickness=0.025,  # For body
     ):
-        super().__init__()
+        super().__init__(name=name)
         if body_half_size:
             self.body_half_size = body_half_size
         else:
@@ -58,10 +59,9 @@ class PotWithHandlesObject(MujocoGeneratedObject):
     def handle_distance(self):
         return self.body_half_size[1] * 2 + self.handle_length * 2
 
-    def get_collision(self, name=None, site=None):
+    def get_collision(self, site=None):
         main_body = new_body()
-        if name is not None:
-            main_body.set("name", name)
+        main_body.set("name", self.name)
 
         for geom in five_sided_box(
             self.body_half_size, self.rgba_body, 1, self.thickness
@@ -228,8 +228,8 @@ class PotWithHandlesObject(MujocoGeneratedObject):
             return ["handle_2"]
         return ["handle_2_c", "handle_2_+", "handle_2_-"]
 
-    def get_visual(self, name=None, site=None):
-        return self.get_collision(name, site)
+    def get_visual(self, site=None):
+        return self.get_collision(site)
 
 
 def five_sided_box(size, rgba, group, thickness):
@@ -332,6 +332,7 @@ class BoxObject(MujocoGeneratedObject):
 
     def __init__(
         self,
+        name,
         size=None,
         size_max=None,
         size_min=None,
@@ -340,6 +341,7 @@ class BoxObject(MujocoGeneratedObject):
         friction=None,
         friction_range=None,
         rgba="random",
+        add_material=False,
     ):
         size = _get_size(size,
                          size_max,
@@ -353,10 +355,12 @@ class BoxObject(MujocoGeneratedObject):
                                                friction_range,
                                                DEFAULT_FRICTION_RANGE)
         super().__init__(
+            name=name,
             size=size,
             rgba=rgba,
             density_range=density_range,
             friction_range=friction_range,
+            add_material=add_material,
         )
 
     def sanity_check(self):
@@ -372,12 +376,12 @@ class BoxObject(MujocoGeneratedObject):
         return np.linalg.norm(self.size[0:2], 2)
 
     # returns a copy, Returns xml body node
-    def get_collision(self, name=None, site=False):
-        return self._get_collision(name=name, site=site, ob_type="box")
+    def get_collision(self, site=False):
+        return self._get_collision(site=site, ob_type="box")
 
     # returns a copy, Returns xml body node
-    def get_visual(self, name=None, site=False):
-        return self._get_visual(name=name, site=site, ob_type="box")
+    def get_visual(self, site=False):
+        return self._get_visual(site=site, ob_type="box")
 
 
 class CylinderObject(MujocoGeneratedObject):
@@ -387,6 +391,7 @@ class CylinderObject(MujocoGeneratedObject):
 
     def __init__(
         self,
+        name,
         size=None,
         size_max=None,
         size_min=None,
@@ -395,6 +400,7 @@ class CylinderObject(MujocoGeneratedObject):
         friction=None,
         friction_range=None,
         rgba="random",
+        add_material=False,
     ):
         size = _get_size(size,
                          size_max,
@@ -408,10 +414,12 @@ class CylinderObject(MujocoGeneratedObject):
                                                friction_range,
                                                DEFAULT_FRICTION_RANGE)
         super().__init__(
+            name=name,
             size=size,
             rgba=rgba,
             density_range=density_range,
             friction_range=friction_range,
+            add_material=add_material,
         )
 
     def sanity_check(self):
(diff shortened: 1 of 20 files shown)


yukezhu:
Not all environments are updated?

```
Traceback (most recent call last):
  File "demo_domain_randomization.py", line 61, in <module>
    control_freq=20,
  File "/Users/yukez/RoboticsSuite/robosuite/environments/base.py", line 22, in make
    return REGISTERED_ENVS[env_name](*args, **kwargs)
  File "/Users/yukez/RoboticsSuite/robosuite/environments/two_arm_lift.py", line 186, in __init__
    camera_depths=camera_depths,
  File "/Users/yukez/RoboticsSuite/robosuite/environments/robot_env.py", line 167, in __init__
    ignore_done=ignore_done,
  File "/Users/yukez/RoboticsSuite/robosuite/environments/base.py", line 96, in __init__
    self._load_model()
  File "/Users/yukez/RoboticsSuite/robosuite/environments/two_arm_lift.py", line 320, in _load_model
    self.pot = PotWithHandlesObject()
TypeError: __init__() missing 1 required positional argument: 'name'
```

robosuite/models/objects/generated_objects.py
```diff
@@ -12,6 +12,7 @@ class PotWithHandlesObject(MujocoGeneratedObject):
 
     def __init__(
         self,
+        name,
```
yukezhu, line 15: can we make `name` as an optional argument and automatically generate a name when the argument is not provided?

robosuite/wrappers/domain_randomization_wrapper.py
```diff
@@ -0,0 +1,62 @@
...
+TODO: support image translation
+TODO: support sensor delay / repeat / dropout
+"""
+import numpy as np
+
+from robosuite.wrappers import Wrapper
+from robosuite.utils.mjmod import TextureModder, LightingModder, MaterialModder, CameraModder
+
+
+class DomainRandomizationWrapper(Wrapper):
```
yukezhu, line 18: Can we provide options to perform more "controllable", e.g., using a subset of modders and specifying the subset of randomization parameters and their ranges?

amandlek: yes - I was hoping to get your input on what the ideal input interface for the wrapper initialization would look like, given the functionality we would like to support (e.g. camera, lighting, and texture modders, and each should have their own parameters for the degree of randomization, how often it happens, and whether it is "global" vs. "local" randomization)

robosuite/models/objects/generated_objects.py
amandlek: I think it is best to explicitly require a name - otherwise we would need to generate names that do not conflict with existing names by using the current system time for example. But I think in most use cases, these objects are created within environments that should be keeping track of their objects regardless

amandlek:
Yes it is listed under the first TODO in the description - I did not what to update all tasks until we agree on the changes looking good for the Lift task.

robosuite/models/objects/objects.py
```diff
@@ -217,12 +219,17 @@ def __init__(
             size ([float], optional): of size 1 - 3
             rgba (([float, float, float, float]), optional): Color
             density (float, optional): Density
-            friction (float, optional): tangentia friction
+            friction (float, optional): tangential friction
```
cremebrule, line 222: since we're overhauling the Object classes, might be good to condense a lot of our interfaces. e.g: having both a `density` and `density_range` option seems a bit redundant and confusing; having a single argument instead that will be interpreted accordingly depending on whether the input is an iterable or single float value could be more efficient

cremebrule: could also be good to either allow adjustment of all friction values, or set the torsional / rolling friction proportional to the tangential friction value

robosuite/scripts/demo_device_control.py
```diff
@@ -151,6 +153,8 @@
         ignore_done=True,
         use_camera_obs=False,
         gripper_visualizations=True,
+        #render_collision_mesh=True,
```
cremebrule, line 156: are these necessary?

robosuite/utils/mjmod.py
```diff
@@ -0,0 +1,935 @@
...
+    well as optimizer parameters like global friction multipliers (eg solimp, solref, etc)
+    To modify a parameteter, use the parameter to be changed as a keyword argument to
+    self.mod and the new value as the value for that argument. Supports arbitray many
+    modifications in a single step.
+    NOTE: It is necesary to perform sim.forward after performing the modification.
+    NOTE: Some parameters might not be able to be changed. users are to verify that
+          after the call to forward that the parameter is indeed changed.
+    """
+    def __init__(self, sim):
+        self.sim = sim
```
cremebrule, line 878: should this extend from `BaseModder`?

robosuite/wrappers/domain_randomization_wrapper.py
```diff
@@ -0,0 +1,218 @@
...
+
+    Dynamics Parameter Noise
+
+        TODO
+
+TODO: Implement Observation Dropout, Action Noise, Dynamics Randomization
+TODO: why can we still see things with all the lights off?
+TODO: each composite object should specify texture groups that get randomized together
+    (geom groups per texture)
+TODO: color perturbations could probably be improved by using a space other than RGB
```
cremebrule, line 57: maybe provide an argument option to specify between HSV or RGB? should only take a couple of transform functions to be added (though propagated through all the classes)

robosuite/wrappers/domain_randomization_wrapper.py
cremebrule: adding a seed option as well could help with deterministic playback and/or benchmarking

robosuite/models/objects/objects.py
amandlek: that's true - but overhaul of Object classes should not be handled in this PR. I think we should discuss what kind of changes we will make to Object classes

robosuite/scripts/demo_device_control.py
amandlek: I actually did not touch this file - I don't know why there is a difference.

robosuite/utils/mjmod.py
amandlek: This is copied from @aprendizaje-de-maquinas - I did not test this yet. I'm not going to handle dynamics randomization in this PR - we can merge that in a separate PR once we decide how to handle it and what kinds to do.

robosuite/wrappers/domain_randomization_wrapper.py
amandlek: I wanted to make sure we have minimal functionality before adding another color space - I think RGB appeared to work fine from my tests. I'd prefer we see a need for other color spaces before we put the effort to include them.

amandlek:
Which TODOs are you referring to?

robosuite/models/objects/objects.py
cremebrule: Looks like we're passing on this for now. I can tackle this in the future with you if you want after we get the v1 essentials up and running. Added this to the nice to have on our doc

cremebrule:
I'm referring to the TODOs in the header docstring of `domain_randomization_wrapper.py`

amandlek:
Yes, those TODOs were my personal notes when implementing the functionality. Most are minor and should be addressed in separate PRs.

robosuite/models/objects/objects.py
```diff
@@ -270,40 +283,71 @@ def get_collision_attrib_template(self):
...
+            "rgb2": array_to_string(self.rgba[:3]), 
+            "width": "100", 
+            "height": "100",
+        }
+
+    def get_material_attrib_template(self):
+        return {
+            "name": "{}_mat".format(self.name), 
+            "texture": "{}_tex".format(self.name), 
+            # "specular": "0.75",
```
roberto-martinmartin, line 301: Why don't we randomize these?

robosuite/scripts/demo_device_control.py
```diff
@@ -101,6 +101,8 @@
 from robosuite import load_controller_config
 from robosuite.utils.input_utils import input2action
 
+import time
```
roberto-martinmartin, line 104: needed?

robosuite/scripts/demo_device_control.py
roberto-martinmartin, line 156: remove comment

robosuite/wrappers/domain_randomization_wrapper.py
roberto-martinmartin: +1 to the seed

robosuite/models/objects/objects.py
amandlek: what do you mean? they are randomized

robosuite/scripts/demo_device_control.py
amandlek: I don't remember touching this file... this is odd. It might be because further updates happened to v1 from the time I branched off to implement this stuff?

robosuite/scripts/demo_device_control.py
amandlek: follow up from above - I can guarantee you that I did not touch this script at all

amandlek: fixed

robosuite/scripts/demo_device_control.py
amandlek: fixed

Issue #76, opened 2020-04-07
Standardize placement initializers for tasks and multiple objects


amandlek:
- Currently, for SawyerPickPlace, the `placement_initializer` argument is ignored, and no placement initializer is used by the PickPlaceTask to place objects - instead the placement is hardcoded such that placement corresponds to the left side of the table (the bin).
- We should unify placement initializers that can work for all tasks across different numbers of objects and different ranges - this will be helpful in addressing issues with SawyerNutAssembly as well where we need different ranges depending on the type of object.

amandlek:
fixed in PR #96

Issue #75, opened 2020-04-03
Add solref and solimp arguments to generated objects


amandlek:
We should add `solref` and `solimp` arguments to `MujocoGeneratedObject` and its subclasses to easily play around and experiment with contact modeling. The default behavior for some objects is pretty bad - for example thin cylinders tend to sink into the table.

amandlek:
fixed in PR #96

Issue #73, opened 2020-04-02
Setting friction for generated objects


amandlek:
The `MujocoGeneratedObject` class supports directly setting a vector of friction coefficients, but subclasses (see [here](https://github.com/StanfordVL/robosuite/blob/master/robosuite/models/objects/objects.py#L247)) but subclasses such as `BoxObject` instead convert the `friction` parameter to a `friction_range` that is just used by the superclass to sample a translational friction value, while keeping the other friction parameters at their default values. Solution is to make sure that the `friction` argument is passed through from subclasses to the super calls.

amandlek:
fixed in PR #96

Issue #66, opened 2020-03-08
placement initializer x and y ranges ignored in SawyerNutAssembly


amandlek:
There is a bug where the x_range and y_range arguments are ignored by the UniformRandomPegsSampler placement initializer - probably because it needs to place multiple object types (square and round nuts) [see here](https://github.com/StanfordVL/robosuite/blob/master/robosuite/models/tasks/placement_sampler.py#L232)

This placement initializer should probably take dictionaries as input to specify ranges per object type and use the input dictionary appropriately.

amandlek:
fixed in PR #96

PR #96, opened 2020-06-11
Placement Initialization and Task Class Improvements

amandlek:
- Before this PR, we had **several** problems with `Task` class organization and how placement sampling worked, especially for the PickPlace and NutAssembly environments. This is because we were unable to handle multi-object placement sampling and initialization, and could not place visual-only objects without changing the Task class.

- Now we have a single `TableTopTask` class responsible for placing normal and visual-only objects. The class reads the `joints` property of each object in order to create the appropriate joints per object, allowing for more flexibility than before.

- We also have a `SequentialCompositeSampler` placement initializer class that can chain multiple placement initializers on top of one another, making multi-object sampling extremely easy.

- I removed the `friction_range` and `density_range` arguments for `MujocoGeneratedObject` instances - dynamics randomization should take place elsewhere. This also fixes an issue where on every reset, objects would sample different densities (masses) by default. For example, on the Lift task, the cube density would vary from 200 to 5000 for any given episode, but now, the cube density default is 1000.

- I also added the option to set `solimp` and `solref` solver parameters for primitive `MujocoGeneratedObject` instances (box, cylinder, ball, and capsule). 

- Finally, I also added a utility script to dump videos of a sequence of sampled environment resets - it is useful for debugging environment initialization. 

- This should close the following issues: #66, #73, #75, #76.


robosuite/environments/two_arm_peg_in_hole.py
```diff
@@ -264,8 +264,9 @@ def _load_model(self):
             self.model.merge(robot.robot_model)
 
         # initialize objects of interest
-        self.mujoco_objects = OrderedDict()
-        self.hole = PlateWithHoleObject()
+        self.hole = PlateWithHoleObject(
+            name="hole",
+        )
         self.cylinder = CylinderObject(
             name="cylinder",
```
yukezhu, line 271: rename to "peg"?

yukezhu: I am a bit confused about the `Task` abstraction. Is `TableTopTask` bound with Table arena? `TwoArmPegInHole` with seems to be an outlier of the TableTopTask. (Also I had no idea from what time the green peg became black after refactoring)

robosuite/scripts/make_reset_video.py
```diff
@@ -0,0 +1,96 @@
+"""
```
yukezhu, line 1: very useful script

robosuite/environments/lift.py
```diff
@@ -319,9 +320,11 @@ def _reset_internal(self):
             # Sample from the placement initializer for all objects
             obj_pos, obj_quat = self.placement_initializer.sample()
 
+            # TODO: clean this up with respect to joint naming?
+
             # Loop through all objects and reset their positions
             for i, (obj_name, _) in enumerate(self.mujoco_objects.items()):
-                self.sim.data.set_joint_qpos(obj_name, np.concatenate([np.array(obj_pos[i]), np.array(obj_quat[i])]))
+                self.sim.data.set_joint_qpos(obj_name + "_0", np.concatenate([np.array(obj_pos[i]), np.array(obj_quat[i])]))
```
yukezhu, line 327: where's the trailing "0" introduced? looks messy that we have to append this "0"  everywhere after the object name

robosuite/environments/two_arm_peg_in_hole.py
amandlek: So I've had nothing to do with the development of the `TwoArmPegInHole` task so re: the naming, perhaps @cremebrule could handle whatever you had in mind here. As for the `TableTopTask` - it is not bound to any one arena - different arenas can be used - you can see that's the case for `PickPlace`.

robosuite/environments/lift.py
amandlek: the reason is that every object gets a free joint by default (see [here](https://github.com/StanfordVL/robosuite/blob/v1_pi/robosuite/models/objects/objects.py#L285)) but more than one joint can be created for an object if desired. Each joint is numbered - take a look at [this](https://github.com/StanfordVL/robosuite/blob/v1_pi/robosuite/models/tasks/table_top_task.py#L89). I agree it's a little messy, but I think it's best to have straightforward naming for the object joints like this.

yukezhu:
In the docstring of `TableTopTask`, it is specified as "... consists of one
robot interacting with a variable number of objects placed on the
tabletop." But the TwoArmPegInHole task clearly doesn't follow this
definition. I wonder if this class encompasses all possible tasks in
robosuite so far? If so shall we use a more generic name for this class?

robosuite/environments/lift.py
yukezhu: I see. I wasn't aware that `obj_name + "_0"` is the name of a joint not an object. This is a little confusing. Maybe we can say `obj_name + "_jnt0"` instead for clarity?

robosuite/environments/door.py
```diff
@@ -304,7 +308,7 @@ def _reset_internal(self):
 
             # # Loop through all objects and reset their positions
             # for i, (obj_name, _) in enumerate(self.mujoco_objects.items()):
-            #     self.sim.data.set_joint_qpos(obj_name, np.concatenate([np.array(obj_pos[i]), np.array(obj_quat[i])]))
+            #     self.sim.data.set_joint_qpos(obj_name + "_jnt0", np.concatenate([np.array(obj_pos[i]), np.array(obj_quat[i])]))
```
cremebrule, line 311: this looks like an automated name generation type thing. Was this intentional or was this just to make the sim work? Also, what is this referencing exactly? From above it seems like the Door object has no joints

robosuite/environments/nut_assembly.py
```diff
@@ -191,12 +191,28 @@ def __init__(
         if placement_initializer:
             self.placement_initializer = placement_initializer
         else:
-            self.placement_initializer = UniformRandomPegsSampler(
-                x_range=[-0.15, 0.],
-                y_range=[-0.2, 0.2],
-                z_range=[0.02, 0.10],
+            # treat sampling of each type of nut differently since we require different
+            # sampling ranges for each
+            self.placement_initializer = SequentialCompositeSampler()
```
cremebrule, line 196: how easy is it to extend this functionality to arena-specific initializations? (e.g: peg locations). Was looking at `SequentialCompositeSampler` and it looks like it can be used pretty flexibly in this way

robosuite/environments/pick_place.py
```diff
@@ -399,31 +462,38 @@ def _load_model(self):
...
+            lst.append((str(self.vis_inits[j]), visual_ob))
+        self.visual_objects = OrderedDict(lst)
 
         lst = []
         for i in range(len(self.ob_inits)):
-            ob = self.ob_inits[i]()
-            lst.append((str(self.item_names[i]) + "0", ob))
+            ob = self.ob_inits[i](
+                name=(self.item_names[i] + "0"),
+                joints=[dict(type="free", damping="0.0005")], # damp the free joint for each object
```
cremebrule, line 476: is this for stability reasons?

robosuite/environments/two_arm_peg_in_hole.py
cremebrule: I haven't touched `TwoArmPegInHole` either, so not sure why the colors have changed. I'll fix it directly on v1.0 branch

cremebrule: what exactly did you want to rename to peg? The cylinder object?

robosuite/models/tasks/table_top_task.py
```diff
@@ -1,36 +1,66 @@
...
+        mujoco_robots, 
+        mujoco_objects, 
+        visual_objects=None, 
+        initializer=None,
+    ):
         """
         Args:
             mujoco_arena: MJCF model of robot workspace
             mujoco_robots: MJCF model of robot model(s) (list)
             mujoco_objects: a list of MJCF models of physical objects
```
cremebrule, line 31: would it make more sense to rename this to collision_objects to differentiate between this and the visual objects?

robosuite/environments/two_arm_peg_in_hole.py
yukezhu: Yes. The task is called peg in hole and the cylinder object is the peg

robosuite/environments/door.py
amandlek: Yes you're right - I just made this modification in case you need to do any resetting for the door here - but we will be abandoning this reset scheme anyways. As for the name - it is automatic name generation for each generated joint per object - take a look at `TableTopTask` - it uses an objects ` joints` property to create desired joints per object, and when creating an object, you can pass a list of `joints` to create the joints for that object. By default a free joint is created per object, but for the door, I explicitly pass an empty list to ensure that no joint is created for the entire door object (the door xml creates hinge joints for the door and the handle).

robosuite/environments/nut_assembly.py
amandlek: I think it would be straightforward to do so - but we don't have to handle that in this PR. I agree in principle it would be nice to have python files per arena that just append fixed object locations (e.g. pegs) and merge cameras into the xml as well, but no need to do that right now.

robosuite/environments/pick_place.py
amandlek: I mirrored this from the `PickPlaceTask` - I think we have always had that damping on the objects for this task

robosuite/models/tasks/table_top_task.py
amandlek: we could certainly do that, but I left it as is for backwards compatibility because we have this `mujoco_objects` dictionary as a property in all environments too. We can certainly change it if both you and @yukezhu think it should be done - let me know

PR #142, opened 2020-12-10
Adding option to select GPU device for offscreen renderer

amandlek:
- All environments now have an argument called `render_gpu_device_id` allowing for selection of GPU device to use for offscreen rendering. This is important for multi-GPU machines.

### DIFF
--- a/robosuite/environments/base.py
+++ b/robosuite/environments/base.py
@@ -80,6 +80,8 @@ class MujocoEnv(metaclass=EnvMeta):
         render_visual_mesh (bool): True if rendering visual meshes
             in camera. False otherwise.
 
+        render_gpu_device_id (int): corresponds to the GPU device id to use for offscreen rendering.
+
         control_freq (float): how many control signals to receive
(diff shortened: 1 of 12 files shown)


robosuite/environments/base.py
```diff
@@ -104,6 +106,7 @@ def __init__(
         render_camera="frontview",
         render_collision_mesh=False,
         render_visual_mesh=True,
+        render_gpu_device_id=-1,
```
cremebrule, line 109: I assume -1 corresponds to no GPU (or maybe default?) -- can we make this explicit in the docstring?

cremebrule (commented):
one small comment, otherwise LGTM

robosuite/environments/base.py
amandlek: yeah it's inferred according to this [here](https://github.com/openai/mujoco-py/blob/master/mujoco_py/mjrendercontext.pyx#L104) - we can update the docstring

Issue #136, opened 2020-12-05
Numpy Linalg Error when training model with PPO algorithm


hermanjakobsen:
Hi! 

I have been training my models using the [PPO](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/ppo/ppo.py) with the `OSC_POSE` controller. Quite rarely the following error occurs 
```
Traceback (most recent call last):
  File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/stable_baselines3/common/vec_env/subproc_vec_env.py", line 29, in _worker
    observation, reward, done, info = env.step(data)
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/stable_baselines3/common/monitor.py", line 97, in step
    observation, reward, done, info = self.env.step(action)
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/robosuite/wrappers/gym_wrapper.py", line 102, in step
    ob_dict, reward, done, info = self.env.step(action)
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/robosuite/environments/base.py", line 280, in step
    self._pre_action(action, policy_step)
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/robosuite/environments/robot_env.py", line 348, in _pre_action
    robot.control(robot_action, policy_step=policy_step)
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/robosuite/robots/single_arm.py", line 260, in control
    torques = self.controller.run_controller()
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/robosuite/controllers/osc.py", line 317, in run_controller
    lambda_full, lambda_pos, lambda_ori, nullspace_matrix = opspace_matrices(self.mass_matrix,
  File "/home/hermankj/Documents/masters_thesis/venv/lib/python3.8/site-packages/numba/np/linalg.py", line 824, in _inv_err_handler
    raise np.linalg.LinAlgError(
numpy.linalg.LinAlgError: Matrix is singular to machine precision.
```
[...]

PR #143, opened 2020-12-10
Update opspace inverses to be pseudoinverses

amandlek:
- Replace some inverses with pseudoinverses to address #136


cremebrule (commented):
Theoretically looks good; is there any quantifiable way we have / can test for robustness using pinv?

amandlek:
How did you test before when changing OSC functionality? I verified that everything works properly using teleoperation - are there other tests you hand in mind? I also tried using my fix here on the data provided in issue #136 and this fixes that particular problem.

PR #144, opened 2020-12-14
Add wipe env tuning adjustments, update gym wrapper to be more robust and compatible

cremebrule:


robosuite/wrappers/gym_wrapper.py
```diff
@@ -35,8 +36,13 @@ def __init__(self, env, keys=None):
...
 
         if keys is None:
-            assert self.env.use_object_obs, "Object observations need to be enabled."
-            keys = ["object-state"]
+            keys = []
+            # Add object obs if requested
+            if self.env.use_object_obs:
+                keys += ["object-state"]
+            # Add image obs if requested
+            if self.env.use_camera_obs:
```
amandlek, line 44: should we allow this? the entire image will get flattened and concatenated - perhaps we should just disallow using GymWrapper with image observations?

cremebrule: good question -- I had to implement this in order to test the PPO + Image Obs for the Wipe env. Do Gym environments not support image-based training natively?

amandlek: wait so for image observations you were using flat observation vectors and MLP policies? usually people use conv layers for image observations

cremebrule: No, it was conv layers, which reshaped the flattened input appropriately

roberto-martinmartin: so this is here because it is only for wiping? I was also surprised of seeing these general changes in this PR. looks good to me, though

cremebrule: it was added specifically because I ran into this issue -- previously, I'd never had to do image-based training with gym-style envs, so I never ran into this problem before

roberto-martinmartin (approved):

PR #151, opened 2021-01-01
Add dynamics randomization

cremebrule:
Adding dynamics randomization via the `DynamicsModder` class in `mjmod.py`. This modder allows mid-sim randomization of the following supported properties, sorted by element group (for more information, please see [Mujoco XML Reference](http://www.mujoco.org/book/XMLreference.html))

#### Opt (Global) Parameters
- `density`: Density of the medium (i.e.: air)
- `viscosity`: Viscosity of the medium (i.e.: air)

#### Body Parameters
- `position`: (x, y, z) Position of the body relative to its parent body
- `quaternion`: (qw, qx, qy, qz) Quaternion of the body relative to its parent body
- `inertia`: (ixx, iyy, izz) diagonal components of the inertia matrix associated with this body
- `mass`: mass of the body

#### Geom Parameters
- `friction`: (sliding, torsional, rolling) friction values for this geom
- `solref`: (timeconst, dampratio) contact solver values for this geom
- `solimp`: (dmin, dmax, width, midpoint, power) contact solver impedance values for this geom

#### Joint parameters
- `stiffness`: Stiffness for this joint
- `frictionloss`: Friction loss associated with this joint
- `damping`: Damping value for this joint

The new `DynamicsModder` follows the same basic API as the other `Modder` classes, and allows per-parameter and per-group randomization enabling. Apart from randomization, this modder can also be instantiated to selectively modify values at runtime. A brief example is given below:

```python
import robosuite as suite
from robosuite.utils.mjmod import DynamicsModder
import numpy as np

# Create environment and modder
env = suite.make("Lift", robots="Panda")
modder = DynamicsModder(sim=env.sim, random_state=np.random.RandomState(5))

# Define function for easy printing
cube_body_id = env.sim.model.body_name2id(env.cube.root_body)
cube_geom_ids = [env.sim.model.geom_name2id(geom) for geom in env.cube.contact_geoms]

def print_params():
    print(f"cube mass: {env.sim.model.body_mass[cube_body_id]}")
    print(f"cube frictions: {env.sim.model.geom_friction[cube_geom_ids]}")
    print()
    
# Print out initial parameter values
print("INITIAL VALUES")
print_params()

# Modify the cube's properties
modder.mod(env.cube.root_body, "mass", 5.0)                                # make the cube really heavy
for geom_name in env.cube.contact_geoms:
    modder.mod(geom_name, "friction", [2.0, 0.2, 0.04])           # greatly increase the friction
modder.update()                                                   # make sure the changes propagate in sim

# Print out modified parameter values
print("MODIFIED VALUES")
print_params()

# We can also restore defaults (original values) at any time
modder.restore_defaults()

# Print out restored initial parameter values
print("RESTORED VALUES")
print_params()
```

Note that the modder already has some sanity checks in place to prevent presumably undesired / non-sensical behavior, such as adding damping / frictionloss to a free joint or setting a non-zero stiffness value to a joint that is normally non-stiff to begin with.

Dynamics randomization has also been added to the `DomainRandomizationWrapper`, and is enabled by default. Running `demo_domain_randomization.py` is another method for demo'ing (albeit an extreme example of) this functionality. All of the new functionality has been heavily documented, and can be found in `mjmod.py`.


roberto-martinmartin:
I can't find the code to randomize orientations. Where is it?

robosuite/models/assets/base.xml
```diff
@@ -1,7 +1,7 @@
 <!-- This is the base xml for all physics simulations. Set global configs here. -->
 <mujoco model="base">
   <compiler angle="radian" meshdir="meshes/" inertiagrouprange="0 0"/>
-  <option impratio="20" cone="elliptic"/>
+  <option impratio="20" cone="elliptic" density="1.2" viscosity="0.00002"/>
```
amandlek, line 4: do you know if this will change the default dynamics from v1.1, and cause action playback to fail for demos collected on that version?

robosuite/utils/mjmod.py
```diff
@@ -1397,85 +1398,521 @@ def bitmap(self):
...
-            return r
+    def update_sim(self, sim):
+        """
+        In addition to super method, reset all internal values
+
+        Args:
+            sim (MjSim): MjSim object
+        """
+        super().update_sim(sim=sim)
+        self.save_defaults()
```
amandlek, line 1678: This doesn't seem to restore the defaults, but just saves them - which is in conflict with the docstring - what's the expected behavior?

robosuite/utils/mjmod.py
```diff
@@ -1397,85 +1398,521 @@ def bitmap(self):
...
+            attr (str): Name of the dynamics parameter to modify. Valid options are self.dynamics_parameters
+            val (int or float or n-array): New value(s) to set for the given dynamics parameter. The type of this
+                argument should match the expected type for the given parameter.
         """
-        for to_mod in kwargs:
-            val = kwargs[to_mod]
+        # Make sure specified parameter is valid, and then modify it
+        assert attr in self.dynamics_parameters, "Invalid dynamics parameter specified! Supported parameters are: {};" \
+                                                 " requested: {}".format(self.dynamics_parameters, attr)
+        getattr(self, f"mod_{attr}")(name, val)
```
amandlek, line 1700: cool trick - consider adding a comment here so it's more clear that this is a way to call one of the below methods by name, based on the variable

amandlek (commented):
looks great! left a few comments

cremebrule:
> I can't find the code to randomize orientations. Where is it?

I am not sure what you mean exactly. Do you mean randomizing body quaternions? You could do `modder.mod('body_name', 'quaternion', new_quat_array)`

robosuite/models/assets/base.xml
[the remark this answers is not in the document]
cremebrule: Yes, this will definitely make a difference, as the default medium values are set to 0 (i.e.: a vacuum environment)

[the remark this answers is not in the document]
cremebrule: We could leave the values defaulted to 0, but it seemed to make more sense to at least have a model for the air for a bit more realism in terms of general drag / friction forces

robosuite/utils/mjmod.py
cremebrule: Hm, that is a good point, perhaps the wording was unclear. By "reset", I mean that the internal default values are updated (via the `save_defaults()` method), since I assumed that if you're calling this method, the old `sim` object was stale and there's now a new simulation instance whose dynamics values need to be kept track of. Is there a better way to convey this?

I don't think it would make sense to actually restore the defaults in sim using this method -- for example, if you're reloading from an XML, and want to update the sim for this modder, it wouldn't necessarily make sense to override the values from the loaded XML, right?

robosuite/utils/mjmod.py
cremebrule: Done!

robosuite/utils/mjmod.py
amandlek: sure - when I think of "reset", I think about resetting to old values, not clearing them and saving the new defaults - so it would be nice to clarify in docstring

cremebrule: Done!

yukezhu:
Looks like the "arm going down" problem still exists when lifting the cube in hand? Shall we use a larger friction term for the joints?

yukezhu:
Do we have a demo script for the new `Modder` class for dynamics randomization?

cremebrule:
> Do we have a demo script for the new `Modder` class for dynamics randomization?

Yeah, the script is shown above in this PR. It's a sub-section of the DynamicsRandomization functionality so I wasn't sure if we should include this as a separate script explicitly -- what do you think @yukezhu ?

yukezhu:
I see. I was not aware that the `demo_domain_randomization.py` includes the dynamics randomization, as our documentation has not been updated: https://robosuite.ai/docs/demos.html#domain-randomization I think this should be highlighted both in our online docs as well as in the header of the demo script - that our domain randomization supports both visuals and dynamics.

amandlek (approved):
LGTM!

PR #157, opened 2021-01-05
Fix tune camera

### LINKED ISSUE #154: Fine tune the custom camera
Hi all, I have a question about the [tune_camera script](https://github.com/ARISE-Initiative/robosuite/blob/master/robosuite/scripts/tune_camera.py). The script works fine with the default fixed cameras:
```bash
camera:  frontview
camera:  birdview
camera:  agentview
camera:  sideview
```
However, when I tune the sawyer [eye-in-hand camera](https://github.com/ARISE-Initiative/robosuite/blob/aa9faf0b760331e1664965d2d86233f2e6a73bbb/robosuite/models/assets/robots/sawyer/robot.xml#L103) or custom camera, it does not work. 
```python
<camera mode="fixed" name="eye_in_hand" pos="0.05 0 0" quat="0 0.707108 0.707108 0" fovy="75"/>
```
It seems that not all cameras are found [here](https://github.com/ARISE-Initiative/robosuite/blob/aa9faf0b760331e1664965d2d86233f2e6a73bbb/robosuite/scripts/tune_camera.py#L44):
```python
cameras = wb.findall("camera")
```
Any ideas to deal with the problems? Thanks!

cremebrule:
Resolve #154, now `tune_camera.py` script can handle both inputted camera tags or simply camera names (will initialize cameras to their default sim location), and can also tune nested cameras as well that don't exist at the top worldbody level.

### DIFF
--- a/robosuite/scripts/tune_camera.py
+++ b/robosuite/scripts/tune_camera.py
@@ -13,6 +13,7 @@ import numpy as np
 
 import robosuite
 import robosuite.utils.transform_utils as T
+from robosuite.utils.mjcf_utils import find_elements, find_parent
 
 # some settings
 DELTA_POS_KEY_PRESS = 0.05      # delta camera position per key press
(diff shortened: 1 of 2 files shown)


roberto-martinmartin (commented):
I have a fundamental question: is adding a new body not changing the dynamics of the robot? This would not affect if the camera is attached to the world but will affect for cameras attached to the end-effector and may break the controller computation.

cremebrule:
> I have a fundamental question: is adding a new body not changing the dynamics of the robot? This would not affect if the camera is attached to the world but will affect for cameras attached to the end-effector and may break the controller computation.

A couple things:

- The body is only added specifically for the purpose of manipulating the camera within this script. The assumed use case is that you tune the camera, then copy the resulting tag and paste it into the xml. There is no added body in sim outside of this script.
- The body is effectively a "dummy" body with negligible mass and inertia (1e-08 for all values), so it should not impact dynamics irregardless

amandlek (approved):
This is great work! LGTM

roberto-martinmartin:
As I understand the first point, the script does not run with the simulation

The second point is unclear. even with negligible mass, it could add a new row-column to the inertia matrix as a new link. If I remember correctly, bodies are links in mujoco, with multiple geoms per body.

But if the use case is to not combine this with manipulation, then there is no problem

Issue #166, opened 2021-01-12
Camera observations overwritten by last camera name


zhuyifengzju:
There is a problem in getting multiple camera observations here:
https://github.com/ARISE-Initiative/robosuite/blob/fb9220491607ead5f1cd5eb25d17626c9ae3756c/robosuite/environments/robot_env.py#L312

The `cam_name` will always be the last  element in `self.camera_names` when those sensor functions are called, and therefore choosing multiple camera views will always get the same image across different views.

quinnliu:
@zhuyifengzju Hi Zhu Yi Feng,

I'm new to this repo && I couldn't help but notice your variable for a camera. I'm very interested in extracting as many camera POVs as possible from the MuJoCo simulation environment. Do you have any advice for how I can access the camera in each time step of the simulation? 

```
sim = MjSim(model)
viewer = MjViewer(sim)
viewer.vopt.geomgroup[0] = 0 # disable visualization of collision mesh

for i in range(10000):
    # get camera_1 output in current time step to process to generate current correct action
    # 1080x1080_pixel_RGB_POV_1 = viewer.get_camera_1() # <======== Anything like this exist?

    # get camera_2 output in current time step
    # 1080x1080_pixel_RGB_POV_2 = viewer.get_camera_2() # <======== Anything like this exist?

    # TODO: action = visual_cognitive_computer.see_and_generate_action(1080x1080_pixel_RGB_POV_1, 1080x1080_pixel_RGB_POV_2)

    sim.data.ctrl[:] = action
    sim.step()
    viewer.render()

PR #171, opened 2021-01-13
Fix a bug of returning same image for different camera views in obs

zhuyifengzju:
Fix the issue #166


cremebrule (approved):
There were a couple bugs, so I went ahead and modularized the functionality to a separate method for cleanliness and clarity

PR #167, opened 2021-01-12
updates for h5py

amandlek:
- change `.value` to `[()]` to address latest h5py module to address #162

Issue #173, opened 2021-01-14
Playback script does not reproduce deterministic results


zhuyifengzju:
Currently the scripts for playing back demonstrations seem to have a problem to reproduce the same trajectories as in the collected demonstrations. So if first collect data using:
`python scripts/collect_human_demonstrations.py --device spacemouse`, and then run  `python scripts/playback_demonstrations_from_hdf5.py --folder DEMONSTRATION_FOLDER --use-actions`, it will gives the following error:
`Traceback (most recent call last):
  File "scripts/playback_demonstrations_from_hdf5.py", line 93, in <module>
    assert(np.all(np.equal(states[j + 1], state_playback)))
AssertionError`

Seems like the deterministic playback is broken somehow.

yukezhu:
Confirmed this issue on my end. Example [demo.hdf5](https://utexas.box.com/s/tjlnuec4e3xzjcoalkk9mq8odn3j3l5c)

State-only playback
![image](https://user-images.githubusercontent.com/1192449/104548186-e4dd7800-55f5-11eb-8b43-51926dc7a894.png)

Action playback
![image](https://user-images.githubusercontent.com/1192449/104548270-11918f80-55f6-11eb-8201-57d1ef60078d.png)

yukezhu:
Also, playback script complained about missing `joint_torques` values in hdf5
https://github.com/ARISE-Initiative/robosuite/blob/master/robosuite/scripts/playback_demonstrations_from_hdf5.py#L83

Issue #162, opened 2021-01-11
h5py Dataset object does not seem to have `value` 


zhuyifengzju:
I think in the latest h5py, Dataset object does not seem to have `value`, which breaks the playback example script. I am not sure if I am missing anything here. I'll create a pull request if this problem is common.

PR #174, opened 2021-01-14
repair deterministic playback

amandlek:
- fix some errors that we had in our scripts and data collection from the v1.0 refactor to enable deterministic action playback
- address #173 and #162

### DIFF
--- a/robosuite/scripts/collect_human_demonstrations.py
+++ b/robosuite/scripts/collect_human_demonstrations.py
@@ -64,23 +64,6 @@ def collect_human_trajectory(env, device, arm, env_configuration):
 
         # Run environment step
         env.step(action)
-
-        if is_first:
-            is_first = False
-
-            # We grab the initial model xml and state and reload from those so that
-            # we can support deterministic playback of actions from our demonstrations.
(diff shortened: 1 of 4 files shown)


yukezhu:
Thanks for the quick repair. @zhuyifengzju should test this PR as well.

yukezhu:
Shall we close PR #167 given this one?

zhuyifengzju (approved):
I tested localy, and this PR fixed the issue. And I think the #167 is also resolved in this PR.

yukezhu (changes requested):
Is this assertion too harsh? Maybe we should allow for some tolerance for small numerical errors in checking the state matching.

https://github.com/ARISE-Initiative/robosuite/blob/e8bf463ab55b5b35090e9686a5ed70b67220f527/robosuite/scripts/playback_demonstrations_from_hdf5.py#L94

Also, tried the replay function with one demonstration collected with the keyboard (download this [demonstration](https://utexas.box.com/s/qjppgbae1zh65k1brzgvnvinnvtwuuzw) file) and this assertion throws exception.

yukezhu (approved):

PR #186, opened 2021-02-08
fix spacemouse import

amandlek:


robosuite/devices/__init__.py
```diff
@@ -1,3 +1,8 @@
 from .device import Device
 from .keyboard import Keyboard
-from .spacemouse import SpaceMouse
+try:
+    from .spacemouse import SpaceMouse
+except ImportError:
+    print("""Unable to load module hid, required to interface with SpaceMouse.\n
+           Only Mac OS X is officially supported. Install the additional\n
+           requirements with `pip install -r requirements-ik.txt`""")
```
yukezhu, line 8: We don't have `requirements-ik.txt` any more, it is replaced by `requirements-extra.txt`. Please change it here as well as in `spacemouse.py`.

yukezhu (changes requested):

robosuite/devices/__init__.py
amandlek: done

yukezhu (approved):

PR #203, opened 2021-03-24
fixed osc position controller bug

snasiriany:
I think the OSC_POSITION controller has incorrect values for the fixed orientation. I believe this PR fixes that issue. I verified that this is a valid rotation matrix -- the determinant is equal to 1, and it was generated using the euler2mat function we have.


snasiriany:
Closing the PR, replicating in another PR.

PR #204, opened 2021-03-24
fixed osc position controller bug [updated]

snasiriany:
New PR, ignore the last identical request. I think the OSC_POSITION controller has incorrect values for the fixed orientation. I believe this PR fixes that issue. I verified that this is a valid rotation matrix -- the determinant is equal to 1, and it was generated using the euler2mat function we have.

### DIFF
--- a/robosuite/controllers/osc.py
+++ b/robosuite/controllers/osc.py
@@ -234,7 +234,7 @@ class OperationalSpaceController(Controller):
                 scaled_delta = self.scale_action(delta)
                 if not self.use_ori and set_ori is None:
                     # Set default control for ori since user isn't actively controlling ori
-                    set_ori = np.array([[0, -1., 0.], [-1., 0., 0.], [0., 0., 1.]])
+                    set_ori = np.array([[0., 1., 0.], [1., 0., 0.], [0., 0., -1.]])
             else:
                 scaled_delta = []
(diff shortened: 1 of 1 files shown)


hermanjakobsen:
To my experience, the "correct" values depends on the type of gripper used, or more precisely, the gripper's reference frame. Have you tested your values with all the different types of grippers?

roberto-martinmartin:
Yes, there are some incongruences between end-effectors / robots. We are unifying so that all frames are aligned

snasiriany:
Thanks @hermanjakobsen and @roberto-martinmartin for the comments! So in that case, will there be a separate PR on unifying so that all frames are aligned? Is there anyone actively working on that task?

cremebrule:
> Thanks @hermanjakobsen and @roberto-martinmartin for the comments! So in that case, will there be a separate PR on unifying so that all frames are aligned? Is there anyone actively working on that task?

Hi @snasiriany , yes I've submitted a PR for this (#213 )

JieFeng-cse:
This perfectly solved my problem! cheers

JieFeng-cse:
This perfectly solved my problem! cheers

yukezhu:
This has been merged into master in PR #213. Thanks.

PR #213, opened 2021-04-12
Standardize EEF Frames

cremebrule:
This PR standardizes EEF frames, such that all grippers have identical conventions for better plug-and-play behavior. This also fixes some nuanced bugs related to orientations, such as issues brought up in #204 .

Specifically, the convention is now as follows:

- EEF Sites are located at the grasping contact point for a given gripper. For parallel jaw / three finger grippers, this is the approximate point where the fingers touch together. For the wiping gripper, this is the bottom surface of the wiper.

- For parallel jaw / three finger grippers, the Z-axis aligns with the prior robot arm's wrist joint rotation axis, and the Y-axis always points perpendicular to the gripper actuation axis

- Default initial qpos for robots have also been standardized, such that all robots will start with their gripper oriented in the same way, regardless of the robot / gripper combination.

**Breaking Changes:** A couple of small, but significant changes have been introduced in this PR, and are described below:

- The wiping gripper EEF site has been moved from the middle of the wiper to the bottom surface of the wiper.

- The `important_sites` property in manipulator robot models used to include `ee`, `ee_x`, `ee_y`, and `ee_z`. Since these sites have now been moved to the gripper files, they no longer exist in the the robot model class but now are in the gripper model class.


hermanjakobsen:
Should the eef properties in `single_arm_env` be modified since the eef sites have been moved to the gripper model? For instance, [_eef_xpos](https://github.com/ARISE-Initiative/robosuite/blob/76c45119ad6e84065ac1e01488a107390f652bbe/robosuite/environments/manipulation/single_arm_env.py#L34) is extracted with `self.sim.data.site_xpos[self.robots[0].eef_site_id]`. Will this still work with the new updates?

amandlek (commented):
LGTM in general, although these changes will cause differences in the model xmls. Can we perhaps bump the robosuite version number up so we can distinguish this change? Also, will this cause differences in the environment dynamics (e.g. replaying old demonstrations)?

yukezhu (approved):

PR #232, opened 2021-05-27
Fixed default control for orientation in OSC

### LINKED ISSUE #230: Can we maintain gripper orientation while using "OSC_POSITION" controller?
I'm working with the simple "panda_lift" env with "OSC_POSITION" controller. I'm trying to collect expert demonstrations through "collect human_demonstrations.py" script. But I often observe that the robot ends up with a gripper orientation which does not support picking up the object although the robot is position-wise close to the object. Please refer to the image below for reference.

The main problem is that the gripper orientation changes continuously while we update "OSC_POSITION". Is there a way to suggest the robot controller to keep the gripper orientation constant (parallel to the table in my case) while still obeying "OSC_POSITION" commands? This would allow the robot to consistently perform the lift task.

gripper_orientation_osc_position

mohanksriram:
The current default value of [set_ori](https://github.com/ARISE-Initiative/robosuite/blob/2a34f81983877852009b64ec3a7eae7efcffe001/robosuite/controllers/osc.py#L237) has -1 determinant which happens to be an improper rotation.

I noticed that the default eef orientation is pointing in positive z axis. If we need the eef to point down, we should
1. First rotate 180 degrees about the x axis (multiply by Rx(180).
2. Then rotate 90 degrees about the z axis (multiply by Rz(90).

Applying these two transformations to an identity matrix results in the new rotation matrix values included in this PR. 
I've verified the behaviour of the new default orientation on all single armed robot types.

Please find the images attached showing the above transformations.

default_gripper_orientation_birdview
default_gripper_orientation_frontview
180degree_rotx_gripper_orientation_frontview
final_gripper_orientation_frontview

Fixes #230.

### DIFF
--- a/robosuite/controllers/osc.py
+++ b/robosuite/controllers/osc.py
@@ -234,7 +234,7 @@ class OperationalSpaceController(Controller):
                 scaled_delta = self.scale_action(delta)
                 if not self.use_ori and set_ori is None:
                     # Set default control for ori since user isn't actively controlling ori
-                    set_ori = np.array([[0, -1., 0.], [-1., 0., 0.], [0., 0., 1.]])
+                    set_ori = np.array([[0., 1., 0.], [1., 0., 0.], [0., 0., -1.]])
             else:
                 scaled_delta = []
         # Else, interpret actions as absolute values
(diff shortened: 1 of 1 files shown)


cremebrule (approved):
LGTM!

PR #236, opened 2021-06-28
Fix Joint Velocity Bug

### LINKED ISSUE #201: What's the unit of the velocities of robot joints?
Hello, today I try to control the velocities of robot joints with `controller_config = load_controller_config(default_controller='JOINT_VELOCITY')`, and I find the default value of the maximum velocity is 1. But I cannot find the specific unit of the velocity in the doc. So the question is what's the unit of the velocity of robot joints, and how can I control the joint with a specific velocity (for example, 180°/s)?

cremebrule:
Fixes #201. Removes the joint velocity observation clipping during the controller computation causing inaccurate results, and increases the default kp gain value for faster convergence.

### DIFF
--- a/robosuite/controllers/config/joint_velocity.json
+++ b/robosuite/controllers/config/joint_velocity.json
@@ -4,7 +4,7 @@
   "input_min": -1,
   "output_max": 0.5,
   "output_min": -0.5,
-  "kp": 0.03,
+  "kp": 3.0,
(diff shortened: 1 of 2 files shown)


yukezhu:
Sorry to get to this late. Is this a breaking change to robot dynamics?

yukezhu (approved):
LGTM

PR #249, opened 2021-08-19
Fix quickstart documentation.

Steve-Tod:
1. `gripper.hide_visualization()` is deprecated.
2. `get_collision` is deprecated and should be replaced by `get_obj`.
3. there's already a free joint that is enabled by default. Adding an
extra one lead to error when `model = world.get_model(mode="mujoco_py")`


yukezhu (approved):
LGTM!

PR #254, opened 2021-09-04
Update robosuite reference list

yukezhu:
Please let me know if I missed any. These papers are pulled from Google Scholar based on the citations.


amandlek (approved):
LGTM - no additional papers come to mind, but we can always do another update later if need be

PR #273, opened 2021-12-18
Add pre-commit black formatter and isort

yukezhu:


pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.black]
+line-length = 120
+target-version = ["py36", "py37", "py38"]
+extend-exclude = "robosuite/((models/assets)|(controllers/config))"
+
+[tool.isort]
+profile = "black"
+line_length = 120
+skip=["__init__.py"]
+py_version = "all"
```
roberto-martinmartin, line 10: we could add the line `filter_files = true`
https://pycqa.github.io/isort/docs/configuration/options.html#filter-files
This will apply the file filters even if a user passes the files manually, e.g., `isort *`

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.black]
+line-length = 120
+target-version = ["py36", "py37", "py38"]
+extend-exclude = "robosuite/((models/assets)|(controllers/config))"
+
+[tool.isort]
+profile = "black"
+line_length = 120
+skip=["__init__.py"]
```
roberto-martinmartin, line 9: why do you do this? is it a good practice?

roberto-martinmartin (commented):
lgtm, just a couple of minor comments

pyproject.toml
yukezhu: Running isort on __init__.py files messed up the importing order of the sub-packages. For example, isort would flip the order of these two import lines [here](https://github.com/ARISE-Initiative/robosuite/blob/master/robosuite/models/robots/__init__.py), causing runtime errors as `RobotModel` is supposed to be imported before importing any `manipulators` class.

pyproject.toml
yukezhu: Done.

roberto-martinmartin:
We should add a couple of lines to our docs in the installation section or somewhere to explain how to activate this

yukezhu:
This should be transparent to the users. There's nothing additional to install, and they won't affect the users unless they are making contributions to the codebase. I added some descriptions in the [contributing guideline](https://github.com/ARISE-Initiative/robosuite/blob/pre-commit/CONTRIBUTING.md).

roberto-martinmartin (approved):
lgtm

PR #286, opened 2022-01-19
Fix ur5e

### LINKED ISSUE #252: Wrong kinematical and dynamical properties for UR5e!!
I am building a simulation with the UR5e robot and I noticed that there is a difference between the data on the company website and the robosuite XML file. I think that in the XML file the data belongs to the UR5 robot and not to the UR5e.

I would appreciate any response to this issue!

website link: 
https://www.universal-robots.com/articles/ur/application-installation/dh-parameters-for-calculations-of-kinematics-and-dynamics/

cremebrule:
Fixes #252 with updated UR5e assets (old assets corresponded to UR5)


cremebrule:
@roberto-martinmartin can you review / approve the PR?

roberto-martinmartin:
@cremebrule , you need to take a functional look at this using the igibson renderer.

This is a picture before your changes:
Screen Shot 2022-03-21 at 3 18 11 PM

This is a picture after your changes:
Screen Shot 2022-03-21 at 3 25 24 PM

Things to fix:
- the shapes are clearly wrong. they have the wrong rotation. This is probably because your obj's are exported using the wrong frame convention (z up/down, x up/down...). Please, check with the old model so that they are similarly looking in meshlab
- there is something wrong with the textures that makes the floor texture to also be wrong. I see this message in the terminal that wasn't there before and is probably hinting at the issue:
`passed in one material ends up overwriting multiple materials`
I would again recommend looking at the old model and other models to see how to do it in the right way.

In general, we want to change as little as possible wrt. the old model that was working fine, just with slightly wrong sizes

Thanks for the effort!

PS: to try yourself, merge master into your branch (you could actually commit this updated version to your branch) and then run: 
`python robosuite/demos/demo_renderers.py --renderer igibson`
selecting the UR5e version

roberto-martinmartin (commented):
See the conversation

cremebrule:
@roberto-martinmartin , fixed the orientation problem! I don't see the floor problem, however, the texture is still weird.

The warning seems to be raised from `mesh_renderer_cpu.py` in igibson, though I wasn't able to infer exactly what is being checked for.

I verified that the rendering looks correct in the NVISII renderer as well, so I'm guessing the issue has to do with the .obj and .mtl file interplay.

Do you know how originally converted the UR5e assets for us? I imported both the original (working texture) .obj file and the new updated one into blender and I can't see a discernable difference. I'm not a blender expert however, so perhaps I need to toggle a specific setting to see some differences. @awesome-aj0123 do you have any ideas?

roberto-martinmartin:
@cremebrule , I'm not sure I understand the current status. Do you load the old UR5e and the new one and they look the same in iGibson? did you compare their mtl's?

Looking at meshes in blender is not the best as it is clever to align correctly. if you use meshlab, it will apply the same convention to both meshes and you may see the orientation error

yukezhu:
Closing this PR as we no longer support iG renderer

PR #288, opened 2022-01-21
Fixed textures for Pick-Place can and cereal box

yukezhu:
Note: potential breaking change! May require regenerating demonstration datasets for vision-based policy training


yukezhu:
Thanks @awesome-aj0123 !

PR #304, opened 2022-03-13
Switch to use SVL PyBullet release version

yukezhu:
The SVL PyBullet version is needed for the iGibson renderer. I also verified that it is compatible with the IK controller. I thus proposed to update our pybullet dependency to this version.


roberto-martinmartin:
The only bad thing is that we would lose updates to Pybullet. We need to update the SVL version manually, which happens in a lower frequency. But I'm not sure how often is pybullet updated.
A good thing of our version is that it removes some of the multiple warning prints from pybullet
Let me know what you think

roberto-martinmartin (commented):
See the conversation

yukezhu:
I actually prefer the manual updates. The current SVL version works for both the IK controller and the iGibson renderer. Using our in-house bullet version gives us full control of compatibility.

roberto-martinmartin:
oh, then sounds good!

PR #307, opened 2022-03-15
Fix sphere geoms with iGibson rendering

s-tian:
- Fix geom type detection and scale for sphere type objects in iGibson parser

### DIFF
--- a/robosuite/renderers/igibson/igibson_utils.py
+++ b/robosuite/renderers/igibson/igibson_utils.py
@@ -80,8 +80,10 @@ def load_object(
 
     if geom_type == "mesh":
         scale = geom_scale
-    elif geom_type in ["box", "sphere"]:
+    elif geom_type == "box": 
         scale = geom_size * 2
+    elif geom_type == "sphere":
+        scale = np.repeat(geom_size * 2, 3)
     elif geom_type == "cylinder":
         scale = [geom_size[0], geom_size[0], geom_size[1]]
     elif geom_type == "plane":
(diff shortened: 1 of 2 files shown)


robosuite/renderers/igibson/igibson_utils.py
```diff
@@ -80,8 +80,10 @@ def load_object(
 
     if geom_type == "mesh":
         scale = geom_scale
-    elif geom_type in ["box", "sphere"]:
+    elif geom_type == "box":
```
roberto-martinmartin, line 83: Do these changes mean that we can create ellipsoids but we can only create cubes with same height/width/depth? Shouldn't we make the same change for the "box" case as it is done for sphere, i.e.,
`scale = np.repeat(geom_size * 2, 3)`
?

s-tian: I think we can create cubes with differing height/width/depth, since `BoxObject` is constructed with 3-dimensional `size`, the `geom_scale` value here will also be length 3 for the "box". But since `BallObject` has a 1-dimensional `size`, we need to repeat it here since iGibson expects a 3-dimensional `scale`. Currently we can't create ellipsoids (only spheres), but perhaps we could add an additional primitive class for ellipsoids with 3-dimensional `size`, or generalize the current `BallObject`.

roberto-martinmartin: got it
sounds good!

roberto-martinmartin (approved):

Issue #321, opened 2022-05-03
"No "site" with name <>" error when accessing "_eef_xmat" properties


misoshiruseijin:
Accessing `_eef_xmat` property in `single_arm_env`  or the `_eef0_xmat` and `_eef1_xmat` properties in `two_arm_env`, causes an error "Value Error: No "site" with name robot0_ee exists." The error is resolved with these changes to `_eef_xmat`, `_eef0_xmat`, and `_eef1_xmat`:

- replacing "ee" with "grip_site" and "right_ee" with "right_grip_site"
- replacing `pf = self.robots[0].robot_model.naming_prefix` with `pf = self.robots[0].gripper.naming_prefix`

cremebrule:
Hi @misoshiruseijin ,

Thanks for finding this bug! Can you submit a PR with the appropriate fixes? Thanks!

cremebrule:
Thanks for fixing this! Closing now.

PR #326, opened 2022-05-11
Fix No Site Error

misoshiruseijin:
Fixes "No Site Named..." error from call to eef_xmat in single_arm_env and two_arm_env (#321)


cremebrule (approved):
LGTM!

PR #353, opened 2022-09-22
Fix sphere geoms with NVISII renderer

s-tian:
Fixes the same issue with getting geom types as #307 for the NVISII renderer.


awesome-aj0123 (approved):

yukezhu:
Merged. Thanks for the fix!

PR #379, opened 2023-03-02
ports gym wrapper to gymnasium

raghavauppuluri13:
for reference: https://gymnasium.farama.org/content/migration-guide/

### DIFF
--- a/robosuite/wrappers/gym_wrapper.py
+++ b/robosuite/wrappers/gym_wrapper.py
@@ -92,9 +92,9 @@ class GymWrapper(Wrapper, gym.Env):
             np.array: Flattened environment observation space after reset occurs
         """
         if seed is not None:
-            try:
+            if isinstance(seed, int):
                 np.random.seed(seed)
-            except:
+            else:
                 TypeError("Seed must be an integer type!")
         ob_dict = self.env.reset()
         return self._flatten_obs(ob_dict), {}
--- a/robosuite/wrappers/gym_wrapper.py
+++ b/robosuite/wrappers/gym_wrapper.py
@@ -95,7 +95,7 @@ class GymWrapper(Wrapper, gym.Env):
             if isinstance(seed, int):
                 np.random.seed(seed)
             else:
-                TypeError("Seed must be an integer type!")
+                raise TypeError("Seed must be an integer type!")
         ob_dict = self.env.reset()
         return self._flatten_obs(ob_dict), {}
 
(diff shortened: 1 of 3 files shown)


robosuite/wrappers/gym_wrapper.py
```diff
@@ -83,15 +84,20 @@ def _flatten_obs(self, obs_dict, verbose=False):
...
+    def reset(self, seed=None, options=None):
         """
-        Extends env reset method to return flattened observation instead of normal OrderedDict.
+        Extends env reset method to return flattened observation instead of normal OrderedDict and optionally resets seed
 
         Returns:
             np.array: Flattened environment observation space after reset occurs
         """
+        if seed is not None:
+            try:
```
amandlek, line 95: the try and except might be overkill - why not a simple `isinstance(seed, int)` check?

raghavauppuluri13: > the try and except might be overkill - why not a simple `isinstance(seed, int)` check?

That's fair! I can simplify it to that.

> also, should seed setting happen in the reset function? I'm not super familiar with the new gymnasium api, but I would think that the seed should be set before constructing the environment, or maybe in the environment init

In the new gymnasium api, they have seed passed in to the `env.reset`. From the docs: "Therefore, [reset()](https://gymnasium.farama.org/api/env/#gymnasium.Env.reset) should (in the typical use case) be called with a seed right after initialization and then never again."

robosuite/wrappers/gym_wrapper.py
```diff
@@ -83,15 +84,20 @@ def _flatten_obs(self, obs_dict, verbose=False):
...
+        Extends env reset method to return flattened observation instead of normal OrderedDict and optionally resets seed
 
         Returns:
             np.array: Flattened environment observation space after reset occurs
         """
+        if seed is not None:
+            if isinstance(seed, int):
+                np.random.seed(seed)
+            else:
+                TypeError("Seed must be an integer type!")
```
amandlek, line 98: you need a `raise` statement here right? After that fix, I think we can merge

amandlek (commented):
We can merge after a minor fix

robosuite/wrappers/gym_wrapper.py
raghavauppuluri13: Sorry, missed that! Should be fixed now

yukezhu (approved):
LGTM

yukezhu:
@raghavauppuluri13 thanks for your PR!

PR #387, opened 2023-03-29
Resolve #386: Reset current_action of grippers at robot reset (SingleArm and Bimanual)

### LINKED ISSUE #386: Gripper current_action not reset at robot.reset
If current_action is not reset to zero at robot resets, the action before the reset has an influence on the action after.

Script to reproduce:

```
import numpy as np

import robosuite

env = robosuite.make(
    "Lift",
    robots="Panda",
    gripper_types="RethinkGripper",
    has_renderer=False,
    has_offscreen_renderer=False,
    use_object_obs=False,
    use_camera_obs=False,
    horizon=100,
    hard_reset=False,  # If True, the problem does not occur because the robot is reloaded
)


# Choose an action with gripper actuation != 0
action = np.zeros(env.action_dim)
action[env.robots[0].controller.control_dim:] = 1

# First test: reset and step once, gripper qpos values should match
np.random.seed(0)
env.reset()
print(f"Current action #1: {env.robots[0].gripper.current_action}")
env.step(action)  # Error becomes apparent after two steps
obs_1, _, _, _ = env.step(action)

np.random.seed(0)
print(f"Current action #2: {env.robots[0].gripper.current_action}")
env.reset()
print(f"Current action #3: {env.robots[0].gripper.current_action}")  # this value should be the same as #1 and different from #2
env.step(action)
obs_2, _, _, _ = env.step(action)

# What actually happens: values do not match due to to the gripper current_action not being reset
if not np.allclose(obs_1["robot0_gripper_qpos"], obs_2["robot0_gripper_qpos"]):
    print(f"gripper qpos values do not match without resetting current_action: {obs_1['robot0_gripper_qpos']} vs {obs_2['robot0_gripper_qpos']}")

# Second test: do the same as above, but manually reset the gripper current_action
np.random.seed(0)
env.reset()
env.robots[0].gripper.current_action = np.array([0.0, 0.0])
env.step(action)
obs_1, _, _, _ = env.step(action)

np.random.seed(0)
env.reset()
env.robots[0].gripper.current_action = np.array([0.0, 0.0])
env.step(action)
obs_2, _, _, _ = env.step(action)

# This time no message is printed, as the gripper qpos values match
if not np.allclose(obs_1["robot0_gripper_qpos"], obs_2["robot0_gripper_qpos"]):
    print(f"gripper qpos values do not match when resetting current_action: {obs_1['robot0_gripper_qpos']} vs {obs_2['robot0_gripper_qpos']}")

```

Sh0cktr4p:
Contributions:
- -Reset gripper current_action at robot reset calls (see issue #386 )

### DIFF
--- a/robosuite/robots/single_arm.py
+++ b/robosuite/robots/single_arm.py
@@ -175,10 +175,12 @@ class SingleArm(Manipulator):
         # First, run the superclass method to reset the position and controller
         super().reset(deterministic)
 
-        if not deterministic:
-            # Now, reset the gripper if necessary
-            if self.has_gripper:
+        # Now, reset the gripper if necessary
+        if self.has_gripper:
+            if not deterministic:
                 self.sim.data.qpos[self._ref_gripper_joint_pos_indexes] = self.gripper.init_qpos
+
+            self.gripper.current_action = np.zeros(self.gripper.dof)
 
         # Update base pos / ori references in controller
         self.controller.update_base_pose(self.base_pos, self.base_ori)
--- a/robosuite/robots/bimanual.py
+++ b/robosuite/robots/bimanual.py
@@ -203,14 +203,15 @@ class Bimanual(Manipulator):
         # First, run the superclass method to reset the position and controller
         super().reset(deterministic)
 
-        if not deterministic:
-            # Now, reset the gripper if necessary
-            for arm in self.arms:
-                if self.has_gripper[arm]:
-                    self.sim.data.qpos[self._ref_gripper_joint_pos_indexes[arm]] = self.gripper[arm].init_qpos
-
         # Setup arm-specific values
         for arm in self.arms:
+            # Now, reset the grippers if necessary
+            if self.has_gripper[arm]:
+                if not deterministic:
+                    self.sim.data.qpos[self._ref_gripper_joint_pos_indexes[arm]] = self.gripper[arm].init_qpos
+
+                self.gripper[arm].current_action = np.zeros(self.gripper[arm].dof)
+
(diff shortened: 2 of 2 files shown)


robosuite/robots/single_arm.py
```diff
@@ -175,11 +175,13 @@ def reset(self, deterministic=False):
...
 
-        if not deterministic:
-            # Now, reset the gripper if necessary
-            if self.has_gripper:
+        # Now, reset the gripper if necessary
+        if self.has_gripper:
+            if not deterministic:
                 self.sim.data.qpos[self._ref_gripper_joint_pos_indexes] = self.gripper.init_qpos
 
+            self.gripper.current_action = np.zeros(self.gripper.dof)
```
amandlek, line 183: does this need to happen if we're not using a deterministic reset? it seems like this will already happen if hard resets are enabled (as issue #386 mentions, the problem is only there when not using hard resets)

amandlek (commented):
Small comment - once resolved, we can merge this

robosuite/robots/single_arm.py
Sh0cktr4p: As the docstring of the method refers to hard resets from XML files as only an example for deterministic resets I thought implementing it like this would make it more clear. However, in practice, `deterministic = True` should exclusively be the case when loading from XML files, is that correct?

amandlek: I see your point - it might be good to refresh it regardless

amandlek (approved):
LGTM

yukezhu:
@Sh0cktr4p Thank you for the PR!

Issue #385, opened 2023-03-25
SpaceMouse class seems to have bugs


tadashiK:
## Environments

- Ubuntu 22.04 with open-source Linux drivers for 3Dconnexion devices
- Using SpaceMouse Compact being detected by hid.enumerate().
- Robosuite installed in Conda's virtual environment
- Custom macro setup with correct device and vendor IDs

## Bug description

I am using SpaceMouse compact and encountered the following issue.

When I run `robosuite/robosuite/demos/demo_device_control.py`, hid fails to open my SpaceMouse Compact even if I set device and vendor IDs correctly in `macros_private.py`. This is weird because I was able to manually open it by `device = hid.device(); device.open(9583, 50741)`.

It turned out that SpaceMouse class uses default values of `vendor_id=macros.SPACEMOUSE_VENDOR_ID` and `product_id=macros.SPACEMOUSE_PRODUCT_ID` imported from `macros.py` instead of `macros_private.py`. If this is an intended behavior, it's fine. However, it is a bit weird to me that robosuite asks users to set up `macros_private.py` but does't use it internally.

Another lines that are seemingly a bug is the following;
```python
        while True:
            d = self.device.read(13)
            if d is not None and self._enabled:

                if macros.SPACEMOUSE_PRODUCT_ID == 50741:
                    ## logic for older spacemouse model
```
These lines check if `SPACEMOUSE_PRODUCT_ID` is 50741, but if SpaceMouse class is instantiated with `product_id=50741` without specifying `SPACEMOUSE_PRODUCT_ID` in `macros.py`, these line should check if `product_id` provided was 50741.

If these are indeed bugs, I will submit a PR to fix them.

tadashiK:
Also, another thing that seems to be a bug is that in both spacemouse and keyboard control cases, pressing ESC does not end the simulation. Indeed, the code does not seem to be detecting whether ESC key is pressed or not.

If you want me to fix this including above, please let me know if those are actually intended or indeed bugs. I can't do anything otherwise.

tadashiK:
I found [robosuite/devices/spacemouse.py](https://github.com/ARISE-Initiative/robosuite/blob/master/robosuite/devices/spacemouse.py) also uses `import robosuite.macros as macros`, which loads the default `macros.py` instead of `private_macros.py`. Maybe the same issue exists in other files too?

snasiriany:
[...]

PR #389, opened 2023-04-07
fix macro issues from issue #385

snasiriany:
* try to import `macros_private.py` inside `macros.py`
* change the default spacemouse product id from `50735` to `50734`
* remove `ESC` option in `keyboard.py` and `spacemouse.py` -- we never have supported this
* fix bug in `spacemouse.py` when referencing spacemouse product id (use the value specified to constructor rather than the macro value)

### DIFF
--- a/robosuite/__init__.py
+++ b/robosuite/__init__.py
@@ -1,13 +1,3 @@
-try:
-    from robosuite.macros_private import *
-except ImportError:
-    import robosuite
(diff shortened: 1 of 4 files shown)


amandlek (approved):
LGTM

zhuyifengzju:
Looks good to me

yukezhu:
Please ping me when you resolve the conflicts. This PR is ready to merge.

PR #391, opened 2023-04-07
logging fixes

snasiriany:
* set unique logging file name: `/tmp/robosuite_{time}_{pid}.log`
* set file logging to `None` by default in `macros.py` (I figured this isn't a critical feature right now)


amandlek (approved):
LGTM. Only comment is that if there is a specific issue that this PR addresses, we should include it in the PR notes

yukezhu (approved):

PR #394, opened 2023-04-12
Add tool hang and transport tasks

snasiriany:
* add missing tasks from robomimic (tool hang and transport)
* ported from `offline_study` branch


yukezhu:
Can the Transport task be accomplished with [1] Single Arms Parallel and [2] Bimanual settings? These are provided as options in `demo_random_actions.py`, but I wasn't sure if we should offer those options.

robosuite/utils/observables.py
```diff
@@ -376,7 +376,8 @@ def _check_sensor_validity(self):
             _ = self.modality
             self._data_shape = np.array(self._sensor({})).shape
             self._is_number = len(self._data_shape) == 1 and self._data_shape[0] == 1
-        except:
+        except Exception as e:
+            print(f"Error: {e}")
```
yukezhu, line 380: Can we use the [logging utils](https://github.com/ARISE-Initiative/robosuite/blob/6fbfae0227fbafe90911a9c57a8f9ed4623df6f0/robosuite/utils/log_utils.py) (DefaultLogger) for error messages?

yukezhu (changes requested):
In general, looks good to me. However, the changes do not follow our coding conventions. Please run our pre-commit hooks `pip install pre-commit; pre-commit install;` and `pre-commit run --all-files;` to check the coding style.

yukezhu (approved):
New changes LGTM

PR #402, opened 2023-05-19
fix Iterable import for python 3.10 in multi_table_arena

### LINKED ISSUE #355: Python 3.10 support
The current numba dependency does not support Python 3.10:

```
  File "/tmp/easy_install-hl2uylwf/numba-0.53.1/setup.py", line 47, in _guard_py_ver
RuntimeError: Cannot install on Python version 3.10.6; only versions >=3.6,<3.10 are supported.
```

Using `"numba>=0.52.0",` and installing from source seems to work, but I haven't fully tested.

minhnh:
- import `Iterable` from the correct module in `robosuite/models/arenas/multi_table_arena.py`
- resolve #355

### DIFF
--- a/robosuite/models/arenas/multi_table_arena.py
+++ b/robosuite/models/arenas/multi_table_arena.py
@@ -1,4 +1,4 @@
-from collections import Iterable
+from collections.abc import Iterable
 
 import numpy as np
 


yukezhu (approved):
LGTM

PR #420, opened 2023-09-07
some minor changes

amandlek:
- fix bug in wrapper
- edit compiler options to enable autolimits (see [this link](https://mujoco.readthedocs.io/en/stable/XMLreference.html#compiler-autolimits) - the default value will change to true in a future MuJoCo release anyways)
- API functionality for composite body

### DIFF
--- a/robosuite/environments/manipulation/pick_place.py
+++ b/robosuite/environments/manipulation/pick_place.py
@@ -68,6 +68,11 @@ class PickPlace(SingleArmEnv):
 
         bin2_pos (3-tuple): Absolute cartesian coordinates of the goal bin
 
+        z_offset (float): amount of z offset for initializing objects in bin
+
+        z_rotation (float, tuple, or None): if provided, controls the range of z-rotation initialization
+            for the objects
+
         use_camera_obs (bool): if True, every observation includes rendered image(s)
 
         use_object_obs (bool): if True, include object (cube) information in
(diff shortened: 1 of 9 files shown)


yukezhu:
@zhuyifengzju FYI- Ajay and I went over the changes in a Zoom call. But we would like to get a second pair of eyes to review this PR. Thanks!

zhuyifengzju:
I've got to review the code here. Overall implementation logic looks good to me. The one question I have here, though, is: why `get_bounding_box_size()` is set with the half-length / radius of a primitive? In that case, probably the function name should be `get_half_bounding_box_size()`.  Can you guys comment on this?  Thanks!

amandlek:
No strong reason why - we just used that convention and stuck with it. We can change the method name if needed

zhuyifengzju:
I think we should change the function name in that case. I think the current naming will cause confusion to people, since a bounding box usually wraps an entire object. My thought is to replace the current `get_bounding_box_size()` to `get_bounding_box_half_size()`, and has a base class function `get_bounding_box_size()` that essentially returns 2x results from `get_bounding_box_half_size()`. With this design, we can keep the current logic, avoid repetitive code, and make them as intuitive as possible to users.

amandlek:
@zhuyifengzju I've gone ahead and implemented your suggestion - let me know if it is good to merge!

zhuyifengzju:
Yep, now it looks all good to me.

zhuyifengzju (approved):

PR #446, opened 2024-01-08
Collect data with fixed freq in real world

chadwick-yao:
Implemented a feature for consistent, real-world frequency data collection in our simulation. This ensures more accurate and reliable data by aligning action execution with realistic time intervals.


zhuyifengzju:
Hi, thanks for the contribution. Before I take a closer look at the file, can you first make it optional in the original data collection script? The current modification also made a lot of changes to the original script that are not critical, which will make it not compatible with previous versions.

zhuyifengzju:
Feel free to re-open this if the required changes are made.

PR #460, opened 2024-03-08
corrected the last link legth of Kinova Gen3

otokatli:
In the [robot.xml](https://github.com/otokatli/robosuite/blob/master/robosuite/models/assets/robots/kinova3/robot.xml) file for Kinova Gen3 robot, there is a minor mistake about the link length of the last link for the robot.

In body `right_hand` the length is defined as 0.65. However, if you check the [Kinova Gen3 manual](https://www.kinovarobotics.com/uploads/User-Guide-Gen3-R07.pdf), on page 199, the link length is defined as 0.615.


amandlek (approved):
LGTM

PR #472, opened 2024-04-11
Add device open fail hint and start control when directly run script

destroy314:


robosuite/devices/spacemouse.py
```diff
@@ -123,7 +126,14 @@ def __init__(
         self.vendor_id = vendor_id
         self.product_id = product_id
         self.device = hid.device()
-        self.device.open(self.vendor_id, self.product_id)  # SpaceMouse
+        try:
+            self.device.open(self.vendor_id, self.product_id)  # SpaceMouse
+        except OSError:
+            print(
+                "Failed to open device. Make sure there are no other "
+                "processes using the device and rerun with sudo."
```
snasiriany, line 134: can you please remove this part about sudo? rerunning the process with sudo is often not the solution we want to recommend to users.

snasiriany (commented):
thank you! left one comment

PR #495, opened 2024-07-09
fix typo: "genereic" to "generic"

hahn5:
There is a typo in `arm_controller_factory` for the joint controllers.


yukezhu (approved):
Thanks!

PR #510, opened 2024-09-05
fix gym_wrapper docstring

bxtbold:
The output type of `reset` of `GymWrapper` didn't match the type on its docstring, so I fixed the docstring instead of removing empty dict (which was gonna result in breaking changes in some demo scripts).


youliangtan (approved):

kevin-thankyou-lin (approved):

kevin-thankyou-lin:
Thanks @bxtbold!

PR #520, opened 2024-09-29
Add function to programmatically set object scale.

johnny-wang16:
Add function to programmatically set object scale. Currently only the following are scaled: geom position, mesh size, body position, site position, slide joint position and slide joint range.


kevin-thankyou-lin:
Thanks @johnny-wang16 ! I'll take a look soon

PR #531, opened 2024-10-30
Fix for test cases

kevin-thankyou-lin:
## What this does

Fixes to pass jaco gripper test, linear interpolation, variable impedance tests.


zhuyifengzju (approved):
The tests all passed on my side. Note that for running test_all_renderes.py on mac, it needs to be on mjpython. Otherwise it will not work. But this is a very specific thing to launch_passive for mac.

snasiriany (approved):
Read thru code, LGTM

PR #543, opened 2024-11-04
Print video save path in demo vid recording

kevin-thankyou-lin:
## What this does
Print video save path in demo vid recording

```python robosuite/demos/demo_video_recording.py```


snasiriany (approved):
LGTM

PR #545, opened 2024-11-04
Update demo docs w/ teleop usage info

kevin-thankyou-lin:
Update demo docs w/ teleop usage info

### DIFF
--- a/docs/demos.md
+++ b/docs/demos.md
@@ -123,6 +123,12 @@ The `demo_device_control.py` scripts shows how to teleoperate robot with [contro
         This current implementation only supports macOS (Linux support can be added).
         Download and install the [driver](https://www.3dconnexion.com/service/drivers.html) before running the script.
 
+* **Mujoco GUI**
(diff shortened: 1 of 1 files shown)


snasiriany (changes requested):
The rest of the docs pertaining to this demo_device_control.py need to be updated:
* example commands have `--controller osc` which is incorrect
* it talks about ik vs osc ("user inputs with ik’s rotations are always taken relative to eef coordinate frame, whereas user inputs with osc’s rotations are taken relative to global frame (i.e., static / camera frame of reference).") which at this point is out of date info and needs to be removed I think.

these are at least the parts I found. Can you please update the PR with all of the changes needed for this section?

kevin-thankyou-lin:
@snasiriany updated

snasiriany (changes requested):
@kevin-thankyou-lin there's a section slightly above that needs to be updated: "The commands are mapped to joint velocities through an inverse kinematics solver from Bullet physics."

kevin-thankyou-lin:
@snasiriany can you merge this?

snasiriany (approved):

snasiriany:
@kevin-thankyou-lin done!

Issue #534, opened 2024-10-30
Demo scripts bugs


youliangtan:
### System Info

```Shell
v1.5
```

### Information

Bug tracker for demo scripts

### Reproduction

1. `python robosuite/demos/demo_control.py`

```sh
Choose a controller for the robot (enter a number from 0 to 4): 3
Traceback (most recent call last):
  File "/home/youliang/robosuite-dev/robosuite/demos/demo_control.py", line 95, in <module>
    options["controller_configs"] = suite.load_part_controller_config(default_controller=controller_name)
  File "/home/youliang/robosuite-dev/robosuite/controllers/parts/controller_factory.py", line 48, in load_part_controller_config
    default_controller in ALL_PART_CONTROLLERS
AssertionError: Error: Unknown default controller specified. Requested WHOLE_BODY_IK, available controllers: ['JOINT_VELOCITY', 'JOINT_TORQUE', 'JOINT_POSITION', 'OSC_POSITION', 'OSC_POSE', 'IK_POSE']
(robosuite-dev) youliang@youliang-Precision-5690:~/robosuite-dev$ 
```

also reported here https://github.com/ARISE-Initiative/robosuite/issues/532

2. `python robosuite/demos/demo_domain_randomization.py`
```sh
  File "/home/youliang/robosuite-dev/robosuite/utils/mjmod.py", line 866, in <listcomp>
    self.textures = [Texture(self.model, i) for i in range(self.model.ntex)]
  File "/home/youliang/robosuite-dev/robosuite/utils/mjmod.py", line 1390, in __init__
    self.tex_rgb = model.tex_rgb
AttributeError: 'MjModel' object has no attribute 'tex_rgb'. Did you mean: 'flex_rgba'?
```

3. `python robosuite/scripts/tune_joints.py`
```
  File "/home/youliang/robosuite_models/../robosuite-dev/robosuite/scripts/tune_joints.py", line 285, in <module>
    controller_config = robosuite.load_controller_config(default_controller="JOINT_POSITION")
AttributeError: module 'robosuite' has no attribute 'load_controller_config'. Did you mean: 'load_part_controller_config'?
```

### Expected behavior

NA

handsomelada:
i have met the same issue

Issue #532, opened 2024-10-30
Bug when update to 1.5.0


Wu-Fisher:
### System Info

```Shell
robosuite                  1.5.0
```

### Information

1. After updating to version 1.5.0, the screen freezes after rendering for a while, but the environment (env) should still be able to step and reset normally.
2. IK controller for part-robot isnt work well and no template config(xx.json) for arm ik control.
3. Some demos do not work. I thk controller module still has some problems.

### Reproduction

1. for any demo

```
for i_episode in range(20):
      ### **next time here off screen will freeze**
        observation = env.reset()
        for t in range(500):
            env.render()
            action = env.action_space.sample()
            observation, reward, terminated, truncated, info = env.step(action)
            if terminated or truncated:
                print("Episode finished after {} timesteps".format(t + 1))
                observation, info = env.reset()
                env.close()
                break
```
2.  Repo didnt contain such file "controllers/config/parts/{}.json"
```
def load_part_controller_config(custom_fpath=None, default_controller=None):
###...
        custom_fpath = os.path.join(
            os.path.dirname(__file__), "..", "controllers/config/parts/{}.json".format(default_controller.lower())
        )

###...
    return controller_config
```

3. python robosuite/demos/demo_control.py >[1] Lift >[3] Panda > [0-4] >
```
  File "/home/wufisher/workspace/dev/robosuite/robosuite/demos/demo_control.py", line 95, in <module>
    options["controller_configs"] = suite.load_part_controller_config(default_controller=controller_name)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wufisher/workspace/dev/robosuite/robosuite/controllers/parts/controller_factory.py", line 48, in load_part_controller_config
    default_controller in ALL_PART_CONTROLLERS
AssertionError: Error: Unknown default controller specified. Requested BASIC, available controllers: ['JOINT_VELOCITY', 'JOINT_TORQUE', 'JOINT_POSITION', 'OSC_POSITION', 'OSC_POSE', 'IK_POSE']
```

kevin-thankyou-lin:
@youliangtan, seems like these tests/demos didn't get fixed in that PR - do you think you could take a stab?

[...]

PR #546, opened 2024-11-04
Fix part controller demo

youliangtan:
### Summary

This addresses both of these issues related to the controller.
 - https://github.com/ARISE-Initiative/robosuite/issues/534
 - https://github.com/ARISE-Initiative/robosuite/issues/532

 - This PR added the previous v1.4 controller configs, so users can still initiate only a single controller (as opposed to `composite), via the `load_part_controller_config()` method.
 - Also there are some inconsistency of using `body_parts` in the internal implementation of handling `controller_configs`, I kept it consistent with the `body_parts_controller_configs` key. Still, I not a fan of dynamically changing the config dict in the internal method.
 - fixes the  `demo_control.py` and `tune_joints.py` scripts.

### What to Run

```sh
 python python robosuite/demos/demo_control.py
```

```sh
 python robosuite/scripts/tune_joints.py 
```

Also all test cases passed.


robosuite/controllers/composite/composite_controller_factory.py
Abhiram824, line 1: Is this change needed? Preferably we could go without this change and just update the entrypoint scripts instead.

robosuite/demos/demo_control.py
Abhiram824, line 1: Ideally, instead of constructing the composite controller config from scratch, we could load the default composite controller for each robot and override the arm controller based on the input to this script

robosuite/controllers/config/parts/default_ur5e.json
Abhiram824, line 1: Do we need this directory: robosuite/controllers/config/ ?

robosuite/controllers/config/parts/default_jaco (copy).json
Abhiram824, line 1: assuming this is not needed?

Abhiram824 (changes requested):
Looks good overall. I think theres some extraneous code in the commit that can be removed though.

robosuite/controllers/composite/composite_controller_factory.py
kevin-thankyou-lin: Do you mean to update `body_parts_controller_configs` to `body_parts` in the `json` files?
I think this more of a style question, but I agree w/ keeping things consistent

robosuite/controllers/config/parts/default_baxter.json
kevin-thankyou-lin, line 1: I think we can remove the  `../config/parts/` directory, since we already have robot defaults in the `robosuite/controllers/config/robots/` folder

robosuite/controllers/config/parts/default_ur5e.json
kevin-thankyou-lin: are you referring to `robosuite/controllers/config/parts/`?

robosuite/demos/demo_control.py
kevin-thankyou-lin: agreed

robosuite/demos/demo_control.py
```diff
@@ -89,10 +89,22 @@
     joint_dim = 6 if options["robots"] == "UR5e" else (16 if options["robots"] == "GR1" else 7)
 
     # Choose controller
-    controller_name = choose_controller()
+    controller_name = choose_part_controller()
+    part_controller_config = suite.load_part_controller_config(default_controller=controller_name)
```
kevin-thankyou-lin, line 93: Could we prompt the user for the actual body part? E.g. humanoid might have multiple body parts and I suspect this script may fail for e.g. GR1FixedLowerBody

kevin-thankyou-lin (changes requested):

kevin-thankyou-lin (approved):

robosuite/utils/input_utils.py
Abhiram824, line 1: This function is no longer needed since `choose_controller(part_controllers=True)` gives the same behavior

robosuite/controllers/composite/composite_controller_factory.py
Abhiram824, line 1: do we need to change the key from `body_parts` to `body_parts_controller_config` I think if don't do this. This PR will have less changes

robosuite/utils/input_utils.py
kevin-thankyou-lin: removed

robosuite/controllers/composite/composite_controller_factory.py
kevin-thankyou-lin: I think updating is fine for consistency?

kevin-thankyou-lin (approved):

Abhiram824 (approved):

kevin-thankyou-lin:
Resolves issue #534

PR #547, opened 2024-11-04
Gymwrapper: support both gym and gymnasium, and support dict_obs and …

youliangtan:
As described in: https://github.com/ARISE-Initiative/robosuite/pull/497 , reopen after v1.5 release

- Support both gymnasium and gym interfaces. A runtime check will ensure gym>=0.26.0 (which uses terminate and truncate).
- Support Flatten and Dict Observation. `GymWrapper(..., flatten_obs=True)`, previously only return flatten obs, which is not always desired.
- Ensure that the `.observation_space` specifies the correct `dtype` and `shape`.

## Quick RUn

```py
import robosuite as suite
from robosuite.wrappers.gym_wrapper import GymWrapper
import cv2

if __name__ == '__main__':
    # create environment instance
    env = suite.make(
        env_name="Lift",  # try with other tasks like "Stack" and "Door"
        robots="Panda",   # try with other robots like "Sawyer" and "Jaco"
        has_renderer=True,
        has_offscreen_renderer=True,
        use_camera_obs=True,
    )

    # convert to gym/gymnasium environment
    env = GymWrapper(
        env,
        keys=["robot0_proprio-state", "object-state", "agentview_image"],
        flatten_obs=False, # default is True, we will not flatten the observation
    )
    print(" -> action space:", env.action_space)
    print(" -> observation space:", env.observation_space)
    obs, info = env.reset()

    for i in range(1000):
        # action = np.random.randn(env.robots[0].dof)  # sample random action
        action = env.action_space.sample()
        obs, reward, done, trunc, info = env.step(action)  # take action in the environment
        print(obs.keys(), obs["agentview_image"].shape)

        # Optional camera view viz
        cv2.imshow("Agent View", cv2.cvtColor(obs["agentview_image"], cv2.COLOR_RGB2BGR))
        if cv2.waitKey(1) & 0xFF == ord('q'): break

        env.render()  # render on display
```

### DIFF
--- a/robosuite/wrappers/gym_wrapper.py
+++ b/robosuite/wrappers/gym_wrapper.py
@@ -4,9 +4,21 @@ This is useful when using these environments with code that assumes a gym-like
 interface.
 """
 
-import gymnasium as gym
 import numpy as np
-from gymnasium import Env, spaces
+
+try:
+    import gymnasium as gym
+    from gymnasium import spaces
+except ImportError:
+    # Most APIs between gym and gymnasium are compatible
+    print("WARNING! gymnasium is not installed. We will try to use openai gym instead.")
+    import gym
+    from gym import spaces
+
+    if not gym.__version__ >= "0.26.0":
+        # Due to API Changes in gym>=0.26.0, we need to ensure that the version is correct
+        # Please check: https://github.com/openai/gym/releases/tag/0.26.0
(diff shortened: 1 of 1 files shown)


kevin-thankyou-lin:
Sorry, just realized that script isn't part of the code base. PR looks good

kevin-thankyou-lin (approved):

PR #548, opened 2024-11-04
Demo fixes

Abhiram824:
## What this does
Updates/Verifies the following demo scripts:
`demo_collect_and_playback_data.py` - fixed but needs the frame rate sleep to make playback at the end smoother. EDIT: this frame rate issue is fixed now

`demo_composite_robot` - most combinations are fine... some off-cases, but overall looks good
`demo_control.py` - runs for all controller types, but erratic behavior when choosing IK
`demo_domain_randomization` - enforced mujoco==3.1.1 and works fine
`demo_device_control.py` - works fine

## How it was tested
ran demo scripts


kevin-thankyou-lin:
To merge after #546  is merged

robosuite/controllers/config/parts/ik_pose.json
kevin-thankyou-lin, line 1: Can we put these into `.../config/default/parts/` folder to be consistent w/ `.../config/default/composite/`?

kevin-thankyou-lin: would need to also update the parts controller_factory.py

robosuite/demos/demo_collect_and_playback_data.py
```diff
@@ -61,10 +61,12 @@ def playback_trajectory(env, ep_dir):
         for state in states:
             env.sim.set_state_from_flattened(state)
             env.sim.forward()
+            env.viewer.update()
```
kevin-thankyou-lin, line 64: could you pull from lastest robosuite master which contains the fps update?

robosuite/demos/demo_device_control.py
```diff
@@ -107,7 +108,12 @@
     parser.add_argument("--arm", type=str, default="right", help="Which arm to control (eg bimanual) 'right' or 'left'")
     parser.add_argument("--switch-on-grasp", action="store_true", help="Switch gripper control on gripper action")
     parser.add_argument("--toggle-camera-on-grasp", action="store_true", help="Switch camera angle on gripper action")
-    parser.add_argument("--controller", type=str, default="BASIC", help="Choice of controller. Can be 'ik' or 'osc'")
+    parser.add_argument(
+        "--controller",
+        type=str,
+        default=None,
+        help="Choice of controller. Can be generic (eg. 'BASIC' or 'WHOLE_BODY_MINK_IK') or json file (see robosuite/controllers/config for examples)",
```
kevin-thankyou-lin, line 115: add the extra option of None to get the robot's default controller if it exists

robosuite/controllers/composite/composite_controller_factory.py
```diff
@@ -18,6 +19,57 @@ def validate_composite_controller_config(config: dict):
             raise ValueError
 
 
+def is_stale_controller_config(config: Dict):
+    """
+    Checks if a controller config is in the format from robosuite versions <= 1.4.1.
```
kevin-thankyou-lin, line 24: I guess this function is now also being used to check if a config is a part-only controller (which we use post v1.4.1 in `demo_control.py`) versus a full composite controller. Maybe we can call it `is_part_controller_config` instead and update the doc string accordingly?

kevin-thankyou-lin (changes requested):

robosuite/controllers/composite/composite_controller_factory.py
Abhiram824: Specifically checks for arm controllers, will rename and update accordingly

kevin-thankyou-lin: I see: I think it's possible to use IK/OSC (and, of course, joint pos/vel/torque) for non-arm controllers

robosuite/controllers/composite/composite_controller_factory.py
```diff
@@ -19,21 +19,21 @@ def validate_composite_controller_config(config: dict):
...
     Args:
         config (dict): Controller configuration
 
     Returns:
-        bool: True if the config is in the old format, False otherwise
+        bool: True if the config is in the for the arm-only, False otherwise
     """
 
-    OLD_CONTROLLER_TYPES = ["JOINT_VELOCITY", "JOINT_TORQUE", "JOINT_POSITION", "OSC_POSITION", "OSC_POSE", "IK_POSE"]
+    ARM_CONTROLLER_TYPES = ["JOINT_VELOCITY", "JOINT_TORQUE", "JOINT_POSITION", "OSC_POSITION", "OSC_POSE", "IK_POSE"]
```
kevin-thankyou-lin, line 34: I think these are basically the part controllers? see message in prev thread

kevin-thankyou-lin:
Look good - thanks!

kevin-thankyou-lin (approved):

PR #552, opened 2024-11-12
Remove unneeded print

kevin-thankyou-lin:
## What this does
Removes unneeded print

### DIFF
--- a/robosuite/robots/fixed_base_robot.py
+++ b/robosuite/robots/fixed_base_robot.py
@@ -49,8 +49,6 @@ class FixedBaseRobot(Robot):
             robot_model=self.robot_model,
             grippers={self.get_gripper_name(arm): self.gripper[arm] for arm in self.arms},
         )
-        print("Loaded composite controller: {}".format(self.composite_controller))
-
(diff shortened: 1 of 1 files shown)


kevin-thankyou-lin:
@youliangtan I removed one print statement to avoid clogging the terminal - could you take a look?

youliangtan (approved):

kevin-thankyou-lin:
Oops @youliangtan looks like you don't have write access --- asking @zhuyifengzju to take a look

zhuyifengzju:
Shall we change it to debug logging messages?

kevin-thankyou-lin:
Looking at the other `.debug()`, these are only used when there's something unexpected e.g.

```
                if part_name not in self._action_split_indexes:
                    ROBOSUITE_DEFAULT_LOGGER.debug(f"{part_name} is not specified in the action space")
                    continue
```

so I think using `.debug()` mightn't fit? Also, not sure how useful logging this is anyways - we only have this print statement for one file

zhuyifengzju (approved):

PR #553, opened 2024-11-13
Add troubleshooting message for spacemouse failure to open

Ynng:
## What this does
Adds helpful error message when SpaceMouse USB device is in use by another process.
Specifically guides users to close `3DconnexionHelper` on macOS, a common cause of device access conflicts.

## How to checkout & try?
1. macOS: Have `3DconnexionHelper` running (reproduces device conflict)
2. Run demo with `--device spacemouse` to see error message

### DIFF
--- a/robosuite/devices/spacemouse.py
+++ b/robosuite/devices/spacemouse.py
@@ -22,6 +22,7 @@ from collections import namedtuple
 
 import numpy as np
 from pynput.keyboard import Controller, Key, Listener
+
 from robosuite.utils.log_utils import ROBOSUITE_DEFAULT_LOGGER
 
 try:
--- a/robosuite/devices/spacemouse.py
+++ b/robosuite/devices/spacemouse.py
@@ -19,6 +19,7 @@ For Linux support, you can find open-source Linux drivers and SDKs online.
 import threading
 import time
 from collections import namedtuple
+from termcolor import colored
 
 import numpy as np
 from pynput.keyboard import Controller, Key, Listener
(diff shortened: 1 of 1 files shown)


robosuite/devices/spacemouse.py
```diff
@@ -128,7 +129,12 @@ def __init__(
         self.vendor_id = vendor_id
         self.product_id = product_id
         self.device = hid.device()
-        self.device.open(self.vendor_id, self.product_id)  # SpaceMouse
+        try:
+            self.device.open(self.vendor_id, self.product_id)  # SpaceMouse
+        except OSError as e:
+            print("Failed to open SpaceMouse device: ", e)
```
kevin-thankyou-lin, line 135: Could we please use `ROBOSUITE_DEFAULT_LOGGER.warning` instead of print statements?

kevin-thankyou-lin (changes requested):
Thanks for your PR @Ynng! Looks great, small comment

robosuite/devices/spacemouse.py
Ynng: for sure, the change has been made and the termcolor import has been removed

kevin-thankyou-lin: thanks!

kevin-thankyou-lin (approved):

kevin-thankyou-lin:
@Ynng can you run `pre-commit run --all-files` to update the formatting? Merging after that

PR #561, opened 2024-11-15
Improve spacemouse

squarefk:
This PR adds support for bimanual teleoperation using the existing SpaceMouse class.
When two space mice are connected, the system will automatically switch to bimanual mode.

To test the functionality without setting up an environment, users can connect either one or two space mice and run `python -m robosuite.devices.spacemouse`.


robosuite/devices/device.py
```diff
@@ -163,10 +163,11 @@ def input2action(self, mirror_actions=False) -> Optional[Dict]:
         )
         ac_dict[f"{active_arm}_abs"] = arm_action["abs"]
         ac_dict[f"{active_arm}_delta"] = arm_action["delta"]
-        ac_dict[f"{active_arm}_gripper"] = np.array([grasp] * gripper_dof)
+        # ac_dict[f"{active_arm}_gripper"] = np.array([grasp] * gripper_dof)
+        ac_dict[f"{active_arm}_gripper"] = [1, 0.45 * grasp, grasp, grasp, grasp, grasp]
```
snasiriany, line 167: what does this part change?

> 0.45 * grasp

this seems specific to the Inspire hand. We should make it a case by case basis
Also not sure if it works for non-humanoids

robosuite/controllers/config/robots/default_gr1_fixed_lower_body.json
```diff
@@ -12,9 +12,9 @@
         "ik_input_rotation_repr": "axis_angle",
         "verbose": false,
         "ik_posture_weights": {
-            "robot0_torso_waist_yaw": 10.0,
-            "robot0_torso_waist_pitch": 10.0,
-            "robot0_torso_waist_roll": 200.0,
+            "robot0_torso_waist_yaw": 30.0,
+            "robot0_torso_waist_pitch": 100.0,
+            "robot0_torso_waist_roll": 100.0,
```
snasiriany, line 17: these changes are making the robot less agile. for example reaching up inside cabinets or farther back in the counter, etc. can we try to make use of the waist only when really needed? I think this part just needs more tuning, perhaps along with the overall gains of the controller.

robosuite/devices/spacemouse.py
```diff
@@ -101,37 +102,26 @@ def convert(b1, b2):
     return scale_to_control(to_int16(b1, b2))
 
 
-class SpaceMouse(Device):
+class SpaceMouseSdkWrapper:
```
snasiriany, line 105: maybe call this SpaceMouseListener instead

robosuite/models/assets/grippers/inspire_left_hand.xml
```diff
@@ -79,12 +79,12 @@
             <site name="ft_frame" pos="0 0 0" size="0.01 0.01 0.01" rgba="1 0 0 0" type="sphere" group="1"/>
             <inertial diaginertia="0 0 0" mass="0" pos="0 0 0"/>
             <body name="eef" pos="0 0 0" quat="0.707 0. -0.707 0.">
-                <site name="grip_site" pos="0 0 0" size="0.01 0.01 0.01" rgba="1 1 0 1" type="sphere" group="2"/>
+                <site name="grip_site" pos="0 0 0" size="0.01 0.01 0.01" rgba="1 1 0 0" type="sphere" group="2"/>
```
snasiriany, line 82: reason for change?

robosuite/models/assets/grippers/inspire_left_hand.xml
```diff
@@ -79,12 +79,12 @@
...
                 <site name="ee_x" pos="0.1 0 0" size="0.005 .1"  quat="0.707105  0 0.707108 0 " rgba="1 0 0 0" type="cylinder" group="1"/>
                 <site name="ee_y" pos="0 0.1 0" size="0.005 .1" quat="0.707105 0.707108 0 0" rgba="0 1 0 0" type="cylinder" group="1"/>
                 <site name="ee_z" pos="0 0 0.1" size="0.005 .1" quat="1 0 0 0" rgba="0 0 1 0" type="cylinder" group="1"/>
                 <!-- This site was added for visualization. -->
-                <site name="grip_site_cylinder" pos="0 0 0" quat="-0.5 -0.5 -0.5 0.5" size="0.005 0.5" rgba="0 1 0 0.3" type="cylinder" group="1"/>
+                <site name="grip_site_cylinder" pos="0 0 0" quat="-0.5 -0.5 -0.5 0.5" size="0.005 0.5" rgba="0 1 0 0" type="cylinder" group="1"/>
```
snasiriany, line 87: reason for change?

robosuite/models/assets/grippers/inspire_left_hand.xml
snasiriany, line 1: can we add some red sites for when each of the other 3 fingers are closed? in a lighter shade of red than the current red color marker of the hand

snasiriany (commented):
Thanks Yu! Left some comments!

snasiriany:
@squarefk also I wanted to press S and start moving the other arm (at the start). but if I do it at the start the other arm does not move. can we fix this?

robosuite/devices/device.py
squarefk: Agreed. This would impact other robots. Moved the changes into class `InspireLeftHand` and `InspireRightHand`

robosuite/devices/spacemouse.py
squarefk: Renamed.

robosuite/models/assets/grippers/inspire_left_hand.xml
squarefk: It has been reverted.

robosuite/models/assets/grippers/inspire_left_hand.xml
squarefk: Make the long ray invisible.

robosuite/models/assets/grippers/inspire_left_hand.xml
squarefk: Done.

squarefk:
> also I wanted to press S and start moving the other arm (at the start). but if I do it at the start the other arm does not move. can we fix this?

Repost the solution here: 

In the is_empty_input_spacemouse function from robocasa/robocasa/scripts/collect_demos.py , we only checked right_delta. Need to update the condition into:

```
    if (
        ("left_delta" not in action_dict or np.all(action_dict["left_delta"] == 0))
        and np.all(action_dict["right_delta"] == 0)
        and action_dict["base_mode"] == -1
        and np.all(action_dict["base"] == 0)
    ):
```

PR #567, opened 2024-11-20
DexMimicGen assets update

xieleo5:
Changes:
* Add _latest_joint_position_actions in WholeBody composite controller class
* Change the action order for inspire hands, now is from pinky to thumb
* Add fourier hands as assets.


robosuite/controllers/composite/composite_controller.py
```diff
@@ -228,6 +228,8 @@ def __init__(self, sim: MjSim, robot_model: RobotModel, grippers: Dict[str, Grip
         # task space actions (such as end effector poses) to joint actions (such as joint angles or joint torques)
 
         self._whole_body_controller_action_split_indexes: OrderedDict = OrderedDict()
+        self._latest_all_joint_angle_action: Optional[np.ndarray] = None
```
kevin-thankyou-lin, line 231: Can we move the joint angle action related code to some different robosuite branch? E.g. maybe v1.5-joint-actions. For this PR, maybe just have the asset/robot modifications?

kevin-thankyou-lin (changes requested):

robosuite/controllers/composite/composite_controller.py
xieleo5: Removed, will create in another PR

kevin-thankyou-lin: perfect

kevin-thankyou-lin (approved):

PR #574, opened 2024-11-21
Fixes #573 🐛

### LINKED ISSUE #573: New PR I made breaks rendering when vram is contested
### System Info

```Shell
I am using Linux ubuntu 24.04 but I suspect this would affect every version of robosuite on all systems.
```

### Information

This problem arises when VRAM is contested and offscreen rendering is used.
I introduced this issue with PR #570

I'll introduce a fix for this momentarily. So sorry to introduce this but hopefully I can get this fix in fast.

### Reproduction

Run a large model and robosuite offscreen rendering at the same time

### Expected behavior

I would expect robosuite rendering to be agnostic to the amount of VRAM available and in use

mishmish66:
## What this does
This fixes issue #573 which I introduced because I passed the argument `EGLConfig * configs` directly instead of by reference to [`eglChooseConfig`](https://registry.khronos.org/EGL/sdk/docs/man/html/eglChooseConfig.xhtml)

## How it was tested
Ran pytest with python 3.11 and 3.12 and used offscreen rendering while running large deep learning tasks

Use a lot of VRAM with pytorch and try offscreen rendering

### DIFF
--- a/robosuite/renderers/context/egl_context.py
+++ b/robosuite/renderers/context/egl_context.py
@@ -126,7 +126,7 @@ class EGLGLContext:
                     "required for creating a headless rendering context."
                 )
             atexit.register(EGL.eglTerminate, EGL_DISPLAY)
-        EGL.eglChooseConfig(EGL_DISPLAY, EGL_ATTRIBUTES, config_ptr, config_size, num_configs)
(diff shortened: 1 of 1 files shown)


mishmish66:
Whil I still think this is a worthwhile change, since it passes args correctly, it seems to make no change to the functionality. I'll continue to try to figure out the cause of #573

PR #584, opened 2024-11-26
Dexmg controller

xieleo5:
Purpose:
Current IK controller will directly apply the solved joint position to robot joints. For some of the robots, we actually want to use joint action space instead of ee_pos space as action to train BC policies. So we need to add a cache that stores intermediate joint actions.

Changes:
1. Add the `_latest_all_joint_angle_action` in `WholeBody` controller to store the latest joint action input
2. Add `_use_joint_angle_action_input` in `WholeBody` to bypass the IK solving step. We can directly pass joint action in `step()`
3. Add `_use_joint_angle_action_input` in configs. 
4. Load `_use_joint_angle_action_input` in `_validate_composite_controller_specific_config`


kevin-thankyou-lin:
This is a do-not-merge branch right?

PR #585, opened 2024-11-28
Update domain randomization wrapper to work with  mujoco!=3.1.1 for some settings

kevin-thankyou-lin:
## What this does
Update domain randomization wrapper to work with  mujoco!=3.1.1 for some settings.
Previously, couldn't use script at all if mujoco!=3.1.1.

## How it was tested
`python robosuite/demos/demo_domain_randomization.py`

### DIFF
--- a/robosuite/wrappers/domain_randomization_wrapper.py
+++ b/robosuite/wrappers/domain_randomization_wrapper.py
@@ -2,6 +2,7 @@
 This file implements a wrapper for facilitating domain randomization over
 robosuite environments.
 """
+import mujoco
 import numpy as np
 
 from robosuite.utils.mjmod import CameraModder, DynamicsModder, LightingModder, TextureModder
@@ -154,6 +155,10 @@ class DomainRandomizationWrapper(Wrapper):
         self.modders = []
 
         if self.randomize_color:
+            assert mujoco.__version__ == "3.1.1", (
+                "TextureModder requires mujoco version 3.1.1 to run. "
+                "Pending support for later versions. Alternatively, you can set randomize_color=False."
+            )
             self.tex_modder = TextureModder(
                 sim=self.env.sim, random_state=self.random_state, **self.color_randomization_args
             )
--- a/robosuite/wrappers/domain_randomization_wrapper.py
+++ b/robosuite/wrappers/domain_randomization_wrapper.py
@@ -5,6 +5,7 @@ robosuite environments.
 import mujoco
 import numpy as np
 
+from robosuite.utils.log_utils import rs_assert
 from robosuite.utils.mjmod import CameraModder, DynamicsModder, LightingModder, TextureModder
 from robosuite.wrappers import Wrapper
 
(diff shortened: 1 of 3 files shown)


robosuite/wrappers/domain_randomization_wrapper.py
```diff
@@ -154,6 +155,10 @@ def __init__(
         self.modders = []
 
         if self.randomize_color:
+            assert mujoco.__version__ == "3.1.1", (
+                "TextureModder requires mujoco version 3.1.1 to run. "
+                "Pending support for later versions. Alternatively, you can set randomize_color=False."
+            )
```
zhuyifengzju, line 161: Let's use the robosuite logger function

zhuyifengzju (changes requested):

robosuite/wrappers/domain_randomization_wrapper.py
kevin-thankyou-lin: `assert robot is not None, "If controller is None, robot must be specified."`

In the repo, there are a lot of examples that use assertions without logging; is a particular reason to use logging here?

zhuyifengzju: I would suggest change those plain printing, instead of following them in the future. It's easier for people to spot the message if there is "[robosuite Error]" etc. People tend to miss those messages within irrelevant track-back messages.

kevin-thankyou-lin: Gotcha, what kind of structure would you use for these logging assertions?

kevin-thankyou-lin: something like:

````python
if [assert condition]:
   robosuite.logger("original assert message")
   assert False, "original assert message"?
```
?

zhuyifengzju: What kind of "structures" are you referring to? You mean like the format inside the brackets?

kevin-thankyou-lin: I mean:

current structure: `assert condition, "message"`

new structure: ?

zhuyifengzju: Oh, sorry, I think I was under some impression.... How about having a function to format the message only? So that we can still have some colored prints. like assert condition, rs_assert_error("message").

kevin-thankyou-lin: sure, one sec

zhuyifengzju: maybe just make it red or any good-looking color to highlight the message will be sufficient.

kevin-thankyou-lin: Updated: now the assert error looks like
image

kevin-thankyou-lin: To get the same error message, set

`        randomize_color=True,  # randomize_color currently only works for mujoco==3.1.1`

in `python robosuite/demos/demo_domain_randomization.py`

kevin-thankyou-lin: @zhuyifengzju does this look good to you?

kevin-thankyou-lin: @zhuyifengzju does this look good to you?

zhuyifengzju (approved):

PR #602, opened 2024-12-30
[fix] fix actuator control with actuator-joint mapping

Dingry:
## What this does
Fix for actuator-to-joint mapping in action handling. Previously, the code didn't properly handle cases where actuator and joint sequences were misaligned or joint count didn't match actuator count due to equality or tendon constraints.

### DIFF
--- a/robosuite/robots/legged_robot.py
+++ b/robosuite/robots/legged_robot.py
@@ -188,6 +188,10 @@ class LeggedRobot(MobileRobot):
             applied_action_high = self.sim.model.actuator_ctrlrange[self._ref_actuators_indexes_dict[part_name], 1]
             actuator_indexes = self._ref_actuators_indexes_dict[part_name]
             actuator_gears = self.sim.model.actuator_gear[actuator_indexes, 0]
+
+            # select only the joints that are actuated
+            actuated_joint_indexes = self._ref_actuator_to_joint_id[actuator_indexes]
+            applied_action = applied_action[actuated_joint_indexes]
             applied_action = np.clip(applied_action / actuator_gears, applied_action_low, applied_action_high)
             self.sim.data.ctrl[actuator_indexes] = applied_action
 
--- a/robosuite/robots/mobile_robot.py
+++ b/robosuite/robots/mobile_robot.py
@@ -266,6 +266,13 @@ class MobileRobot(Robot):
             self.sim.model.joint_name2id(joint) for joint in self.robot_model.head_joints
         ]
 
+        self._ref_actuator_to_joint_id = np.ones(self.sim.model.nu).astype(np.int32) * (-1)
+        for part_name, actuator_ids in self._ref_actuators_indexes_dict.items():
+            self._ref_actuator_to_joint_id[actuator_ids] = np.array([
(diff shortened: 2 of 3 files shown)


robosuite/robots/mobile_robot.py
```diff
@@ -266,6 +266,15 @@ def setup_references(self):
...
 
+        self._ref_actuator_to_joint_id = np.ones(self.sim.model.nu).astype(np.int32) * (-1)
+        for part_name, actuator_ids in self._ref_actuators_indexes_dict.items():
+            self._ref_actuator_to_joint_id[actuator_ids] = np.array(
+                [
+                    self._ref_joints_indexes_dict[part_name].index(self.sim.model.actuator_trnid[i, 0])
+                    for i in actuator_ids
+                ]
+            )
+
```
snasiriany, line 277: I think we should have the same code in `fixed_base_robot.py` as well, right?

robosuite/robots/mobile_robot.py
```diff
@@ -266,6 +266,15 @@ def setup_references(self):
             self.sim.model.joint_name2id(joint) for joint in self.robot_model.head_joints
         ]
 
+        self._ref_actuator_to_joint_id = np.ones(self.sim.model.nu).astype(np.int32) * (-1)
+        for part_name, actuator_ids in self._ref_actuators_indexes_dict.items():
+            self._ref_actuator_to_joint_id[actuator_ids] = np.array(
```
snasiriany, line 271: add a very small comment in the code explaining how this mapping works (conceptually at a high level)

robosuite/robots/legged_robot.py
```diff
@@ -188,6 +189,17 @@ def control(self, action, policy_step=False):
...
+            part_controllers = self.composite_controller.get_controller(part_name)
+            if (
+                isinstance(part_controllers, JointPositionController)
+                or isinstance(part_controllers, JointVelocityController)
+                or isinstance(part_controllers, JointTorqueController)
+            ):
+                # select only the joints that are actuated
+                actuated_joint_indexes = self._ref_actuator_to_joint_id[actuator_indexes]
+                applied_action = applied_action[actuated_joint_indexes]
+
```
snasiriany, line 202: would this logic apply more generally to other robots? other than legged_robot.py

also, why is the logic only applied for JointPositionController/JointVelocityController/JointTorqueController? Is it because these are the controllers where we directly output joint targets?

snasiriany (changes requested):
Left some comments and some questions!

robosuite/robots/legged_robot.py
Dingry: Yes, this logic should apply to all robots.

Yes. The JointController generates control signals for all joints, while the GripperController only produces control signals for actuators. This difference explains why we need these conditional statements. I'm uncertain about the behavior of other controllers. 
Also I was thinking that the GripperController seems essentially similar to JointController but focuses only on actuator-driven joints. If so, we could potentially unify their implementation since their fundamental problem if the same: align joint with actuator actions.

robosuite/robots/mobile_robot.py
Dingry: Yes.

PR #622, opened 2025-01-17
Replace warn with warning

squarefk:
Fix the the following warning message:

robosuite/robosuite/robots/robot.py:151: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead


kevin-thankyou-lin (approved):
Thanks!

PR #624, opened 2025-01-22
fix the mounting direction of dex hand for panda robot

xieleo5:
## What this does
Change the mounting direction for PandaDexHR and PandaDexHL, previously the fingers are pointing forward. Now the hand is pointing downward.

### DIFF
--- a/robosuite/models/robots/compositional.py
+++ b/robosuite/models/robots/compositional.py
@@ -87,7 +87,7 @@ class PandaDexRH(Panda):
 
     @property
     def gripper_mount_quat_offset(self):
-        return {"right": [0.5, -0.5, 0.5, 0.5]}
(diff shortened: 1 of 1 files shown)


xieleo5:
Previous
image

Now
image

snasiriany:
@kevin-thankyou-lin looks like this is a breaking change. Should we allow for backwards compatibility? I would opt for skipping that and keeping things simple.

xieleo5:
If we want to keep backward compatible, I can also add two other robots with the new mounting quat offset. What do you think @snasiriany

kevin-thankyou-lin (approved):

kevin-thankyou-lin:
Looks good to me; there're no datasets tied to this robot so should be fine

PR #626, opened 2025-01-23
[fix] fix mink IK with delta input type

Dingry:
fix frame transformation for delta action input with both base & world frame reference. The fixed robot movement is shown in the video. Left: base, delta; Right: base, absolute. Under absolute mode, the robot will move faster. Not sure if there is an action scaling for this mode.

https://github.com/user-attachments/assets/c2aacab1-e964-4af4-b0ce-89b201caab7b


kevin-thankyou-lin (approved):
thanks!

PR #628, opened 2025-01-24
fix pot with handles when pot is rectangle

xieleo5:
The current pot_with_handles object will result in a wrong shape when `size[0] != size[1]` One line change can fix this.


xieleo5:
Code to test
```
pot = PotWithHandlesObject(
            name="pot",
            body_half_size=[0.16, 0.20, 0.05],
            handle_radius=0.01,
            handle_length=0.06,
            handle_width=0.14,
)
```
Expected behavior, return a rectangle tray with handles.

Before
image
After
image

kevin-thankyou-lin (approved):

PR #631, opened 2025-02-01
Added XArm7

EdwardoSunny:
## What this does
Added the UFactory XArm7 with gripper using files from [mujoco-menagerie](https://github.com/google-deepmind/mujoco_menagerie)

Examples:
|  Title               | Label           |
|----------------------|-----------------|
| Adds xarm7 and xarm7 gripper   | (✨ Feature)     |

image

## How it was tested
Tested using the code inside of `scripts/` to ensure compatibility with robosuite and functionality. If anyone wants to do further testing, simply select the robot as `XArm7` and run program as usual.

Any regular robosuite program, just replace robot with `XArm7`.
``` python
import numpy as np
import robosuite as suite

# create environment instance
env = suite.make(
    env_name="Lift", # try with other tasks like "Stack" and "Door"
    robots="XArm7",  # try with other robots like "Panda", "Sawyer", and "Jaco"
    has_renderer=True,
    has_offscreen_renderer=False,
    use_camera_obs=False,
)

# reset the environment
env.reset()

for i in range(1000):
    action = np.random.randn(*env.action_spec[0].shape) * 0.1
    obs, reward, done, info = env.step(action)  # take action in the environment
    env.render()  # render on display
```

@kevin-thankyou-lin @zhuyifengzju


abhihjoshi (changes requested):
Hi @EdwardoSunny, thanks for creating a PR for this! I've noticed there are a lot of files that have been changed. Do you mind pushing a commit so that only the files that need to be changed show as changed? This might just be a formatting thing.

EdwardoSunny:
> Hi @EdwardoSunny, thanks for creating a PR for this! I've noticed there are a lot of files that have been changed. Do you mind pushing a commit so that only the files that need to be changed show as changed? This might just be a formatting thing.

Yes sorry about that. It must have been the precommit black formatting from the contribution guide. I will fix that and submit a more clean commit. Thanks :)

PR #638, opened 2025-02-18
Enable instance segmentation to include bodies in arena

kevin-thankyou-lin:
## What this does
Enable instance segmentation to include bodies in arena.

Before fix:

Screenshot from 2025-02-18 16-12-22

After fix:

image

## How it was tested
`python robosuite/demos/demo_segmentation.py --segmentation-level instance --random-colors --env-name NutAssemblySquare --camera agentview`

### DIFF
--- a/robosuite/models/tasks/task.py
+++ b/robosuite/models/tasks/task.py
@@ -1,9 +1,39 @@
+import xml.etree.ElementTree as ET
 from copy import deepcopy
+
+import mujoco
 
 from robosuite.models.objects import MujocoObject
 from robosuite.models.robots import RobotModel
 from robosuite.models.world import MujocoWorldBase
 from robosuite.utils.mjcf_utils import get_ids
+
+
+def get_subtree_geom_ids_by_group(model: mujoco.MjModel, body_id: int, group: int) -> list[int]:
+    """Get all geoms belonging to a subtree starting at a given body, filtered by group.
+
+    Args:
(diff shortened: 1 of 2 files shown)


robosuite/models/tasks/task.py
```diff
@@ -106,15 +136,32 @@ def generate_id_mappings(self, sim):
         for robot in self.mujoco_robots:
             models += [robot] + robot.models
 
+        worldbody = self.mujoco_arena.root.find("worldbody")
+        exclude_bodies = ["table"]
+        top_level_bodies = [
```
Abhiram824, line 141: Not sure if its the best idea to just blindly include the top level bodies. I think it would make more sense to pass into the Task constructor the bodies who's geoms you want to have a mask?

Abhiram824 (commented):
Left a comment. Overall, not sure if it makes more sense to enable instance segmentation to include bodies or just create a dummy object class for the pegs and pass those in. If the former is preferred I think these bodies should be passed into the constructor of Task.

Also tested it and it works on my end

robosuite/models/tasks/task.py
kevin-thankyou-lin: Hmm, the main issue here is that if we add dummy objects, they'll still go through e.g. `merge_objects()` and get added. Would need a hacky flag that says to not merge the object, but also unclean.

Abhiram824 (approved):

PR #642, opened 2025-02-24
Update NutAssembly to handle different mounts aside from default

kevin-thankyou-lin:
## What this does

Update NutAssembly to handle different mounts aside from default

## How it was tested

`python robosuite/demos/demo_control.py`, then press 2 3 4


abhihjoshi (approved):

kevin-thankyou-lin:
CC @amandlek

PR #662, opened 2025-03-24
Adding shell inertia for meshes in Sawyer

abhihjoshi:
Add shell inertia to small meshes in Sawyer robot.


kevin-thankyou-lin (approved):
Thanks!

PR #670, opened 2025-04-04
[add] add grasp qpos for inspire and fourier hand

Dingry:
## What this does
- Adding grasping qpos for Inspire and Fourier hands. Used it when treating the hand as a gripper, only control its open/close pose.
- Using this qpos to set the hand action cmd in the device control class.

### DIFF
--- a/robosuite/devices/device.py
+++ b/robosuite/devices/device.py
@@ -112,6 +112,7 @@ class Device(metaclass=abc.ABCMeta):
 
         # Get controller reference
         controller = robot.part_controllers[active_arm]
+        gripper = robot.gripper[active_arm]
         gripper_dof = robot.gripper[active_arm].dof
 
         assert controller.name in ["OSC_POSE", "JOINT_POSITION"], "only supporting OSC_POSE and JOINT_POSITION for now"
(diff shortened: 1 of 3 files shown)


kevin-thankyou-lin:
Thanks! Could you please give a script as example usage too?

Dingry:
`python robosuite/scripts/collect_human_demonstrations.py --robots GR1FixedLowerBody` pressing `s` to test the open/close behavior of the gripper.

previous hand closing qpos v.s. current hand closing qpos: 
image

image

kevin-thankyou-lin:
Thanks! So for these two hands that now have `grasp_qpos`, how does one go back to the non-binary hand action mode? Would one need to comment out the `grasp_qpos` property in code?

Dingry:
Well, I walked through the code. It seems that Robosuite does not support the non-binary control of dexterous hands. I confirmed with @xieleo5 that for multi-finger control, they wrote a new device control class in another repo.

kevin-thankyou-lin:
Gotcha, I think this change works for now. Seems like it'd affect future devices that have non-binary control, however. Those devices would need to override the `input2action` function command.

kevin-thankyou-lin:
would it be possible to make binary grasp opt in rather than opt out? One of the features we want to support in the near future (especially since dexterous hands are getting good) would be non-binary control

kevin-thankyou-lin (approved):

PR #680, opened 2025-04-13
Update mobile_robot.py

ShahRutav:
fix: correct arm value in sensor functions

## What this does
End effector to base sensor functions was using `arm` value by reference (see the function is declared inside the loop). Essentially, if the robot had two arms in the order ["left", "right"], it still returns both values from the right arm.

|  Title               | Label           |
|----------------------|-----------------|
| Fixes        | (🐛 Bug)        |

```bash
python test_arm_reference.py
```

### Test Function

```python
def create_functions():
    functions = []
    arms = ['left', 'right']

    for arm in arms:
        def print_arm(arm=arm):
            print(arm)
        functions.append(print_arm)

    return functions

def create_functions_orig():
    functions = []
    arms = ['left', 'right']

    for arm in arms:
        def print_arm():
            print(arm)
        functions.append(print_arm)

    return functions

print("as an argument")
# Create the functions
funcs = create_functions()

# Call the functions
for func in funcs:
    func()

print("not as an argument")
# Create the functions
funcs = create_functions_orig()

# Call the functions
for func in funcs:
    func()

```

### DIFF
--- a/robosuite/robots/mobile_robot.py
+++ b/robosuite/robots/mobile_robot.py
@@ -322,7 +322,7 @@ class MobileRobot(Robot):
         for arm in self.arms:
 
             @sensor(modality=modality)
-            def base_to_eef_pos(obs_cache):
+            def base_to_eef_pos(obs_cache, arm=arm):
                 eef_pos = np.array(self.sim.data.site_xpos[self.eef_site_id[arm]])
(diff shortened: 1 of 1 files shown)


kevin-thankyou-lin (approved):
Thanks

PR #697, opened 2025-05-13
update part controller config to include joint indices

### LINKED ISSUE #673: Gripper Controller Loading
### System Info

```Shell
v1.5, Ubuntu
```

### Information

_No response_

### Reproduction

In robosuite/robots/robot.py, line 911, the key "joints" is incorrectly loaded when the Panda gripper is selected with the Joint Position controller. The value should be the integer ID of the gripper joints, but the string names of the joints are passed instead. 

```
                self.part_controller_config[gripper_name]["joint_indexes"] = {
                    "joints": self.gripper_joints[arm],
                    "actuators": self._ref_joint_gripper_actuator_indexes[arm],
                    "qpos": self._ref_gripper_joint_pos_indexes[arm],
                    "qvel": self._ref_gripper_joint_vel_indexes[arm],
                }
```

### Expected behavior

_No response_

Abhiram824:
## What this does
Fixes #673 

## How it was tested
To test change the basic controller to use JOINT_POSITION for grippers. (change the this [line](https://github.com/ARISE-Initiative/robosuite/blob/master/robosuite/controllers/config/default/composite/basic.json#L46) to `"type": "JOINT_POSITION"`). Then run `python robosuite/scripts/collect_human_demonstrations.py --environment Lift`

Before the changes in this branch, the script will have an exception. After the changes controlling the gripper with JOINT_POSITION control should work fine.

### DIFF
--- a/robosuite/robots/robot.py
+++ b/robosuite/robots/robot.py
@@ -912,7 +912,7 @@ class Robot(object):
                 self.part_controller_config[gripper_name]["ndim"] = self.gripper[arm].dof
                 self.part_controller_config[gripper_name]["policy_freq"] = self.control_freq
                 self.part_controller_config[gripper_name]["joint_indexes"] = {
-                    "joints": self.gripper_joints[arm],
+                    "joints": self._ref_joints_indexes_dict[gripper_name],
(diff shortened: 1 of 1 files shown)


kevin-thankyou-lin (approved):
LG

PR #701, opened 2025-05-14
Fix: Prevent GPU memory leak by ensuring sim destruction on reset

kywch:
Hello, thank you for maintaining the awesome robosuite!

## What this does

I've experienced a GPU memory leak while running a long session with many resets. I found that `self._destroy_sim()` was not being called during reset because my renderer was `mjviewer`.

With `self._destroy_sim()` properly called in each reset regardless of the renderer, I no longer see the GPU memory leak.

## How it was tested
I monitored `nvidia-smi` during my script before and after the fix.

Thanks!

### DIFF
--- a/robosuite/environments/base.py
+++ b/robosuite/environments/base.py
@@ -274,7 +274,8 @@ class MujocoEnv(metaclass=EnvMeta):
         if self.hard_reset and not self.deterministic_reset:
             if self.renderer == "mujoco":
                 self._destroy_viewer()
-                self._destroy_sim()
+
+            self._destroy_sim()
             self._load_model()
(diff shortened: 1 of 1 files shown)


kywch:
Here is a simple test script to demonstrate the problem. With the current code, memory usage fluctuates up to 2GB. With the fix, memory usage remains constant at around 260MB.

```python
import robosuite

env = robosuite.make("Lift", robots=["Panda"], has_offscreen_renderer=True)

while True:
    env.reset()
    sim_ref = env.sim
    for _ in range(100):
        env.step([0] * 7)
```

Thanks!

PR #736, opened 2025-07-18
Fixes mobile base xmls and controller logic

kevin-thankyou-lin:
Currently, for mobile base xmls, "forward" means "side" while "side" means forward. Consequently `mobile_base/joint_vel.py` has a mysterious action switching logic.

## What this does
Fix mobile base xml so forward means forward (x axis), side means sid…e (y-axis); fix controller's weird x-y axis switching logic.

## How it was tested
`python robosuite/scripts/collect_human_demonstrations.py --robots PandaOmron` still works as expected.


Abhiram824:
Changes look good, and I tested with GR1FloatingBase, PandaOmron, and Tiago and they all worked fine. But didnt we want to maintain some sort of backwards compatibility with this PR or no? I recall @snasiriany wanted this right?

Abhiram824 (approved):

PR #738, opened 2025-07-22
fix site scaling to get all sites from world body

Abhiram824:
## What this does
fixes https://github.com/robocasa/robocasa/issues/154.

The latest commit that refactors `set_scale` sets the scale for sites under the `self._obj` however, it should be looking for sites under `self.worldbody` (as done prior to the latest commit in master commit)

## How it was tested
`python -m robocasa.demos.demo_kitchen_scenes` using public main of robocasa

run above command with this branch and with master. With this branch it should work, on master will get error 

`ValueError: Overhang value is too large, must be lower than front padding (-0.22)`

### DIFF
--- a/robosuite/models/arenas/arena.py
+++ b/robosuite/models/arenas/arena.py
@@ -155,6 +155,7 @@ class Arena(MujocoXML):
         scale_mjcf_model(
             obj=obj,
             asset_root=self.asset,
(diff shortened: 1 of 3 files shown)


kevin-thankyou-lin (approved):
lgtm

PR #751, opened 2025-09-11
Fix binding_utils to work on windows

alek5k:
## What this does
mujoco.dll is not distributed with robosuite. This change looks for the mujoco dll in the mujoco site-packages, and sets the correct MUJOCO_GL environment variable

## How it was tested
Tested on my Windows 11 machine.


kevin-thankyou-lin (approved):
thanks!

PR #758, opened 2025-10-13
[Draft] Support G1 Locomanip

xieleo5:
## What this does
1. Add a external torque compensation flag, enables user to pass torque. This feature can better align sim and real envs
2. Support timestep in observables to make sure playback not diverge
3. Add qacc for robot joints to support WBC, add imu sensor to support WBC
## How it was tested
The G1 locomanip repo will be open-sourced soon. It will include unit test for these features.

### DIFF
--- a/robosuite/controllers/parts/generic/joint_pos.py
+++ b/robosuite/controllers/parts/generic/joint_pos.py
@@ -175,8 +175,8 @@ class JointPositionController(Controller):
         # initialize
         self.goal_qpos = None
 
+        self.desired_torque_as_acceleration = kwargs.get("desired_torque_as_acceleration", True)
         self.use_torque_compensation = kwargs.get("use_torque_compensation", True)
-
         self.use_external_torque_compensation = kwargs.get("use_external_torque_compensation", False)
 
     def set_goal(self, action, set_qpos=None):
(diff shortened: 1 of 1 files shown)


robosuite/controllers/parts/generic/joint_pos.py
```diff
@@ -259,13 +263,13 @@ def run_controller(self):
 
         position_error = desired_qpos - self.joint_pos
         vel_pos_error = -self.joint_vel
-        desired_torque = np.multiply(np.array(position_error), np.array(self.kp)) + np.multiply(vel_pos_error, self.kd)
```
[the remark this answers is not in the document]
xieleo5: @Dingry  just made a fix for this. The idea is that previously, robosuite's "torque" is actually acceleration. But when calculating grav compensation, we want it to be the actual torque. @Dingry just added an extra flag that, by default, still treats torque as acceleration to keep backward compatibility.

robosuite/controllers/parts/controller.py
```diff
@@ -122,6 +122,9 @@ def __init__(
         # Torques being outputted by the controller
         self.torques = None
 
+        # External torque compensation
+        self.external_torque_compensation = 0
```
Abhiram824: Sorry if I am missing something, but right now it looks like there isnt any way to pass in any custom value for `external_torque_compensation`  right?

Abhiram824: But how will you set the value of the flag? right now it seems like there is no way to pass in any values to the class?

PR #784, opened 2025-12-24
Update Python version to 3.11 in workflow

kevin-thankyou-lin:
## What this does
Update Python version to 3.11 in workflow to fix github deploy docs failing.


abhihjoshi (approved):
lgtm

PR #792, opened 2026-01-15
Robocasa dev

Abhiram824:
## What this does

Changes to make robosuite compatible with latest updates to robocasa features

### DIFF
--- a/robosuite/controllers/parts/mobile_base/joint_vel.py
+++ b/robosuite/controllers/parts/mobile_base/joint_vel.py
@@ -300,20 +300,21 @@ class MobileBaseJointVelocityController(MobileBaseController):
     def name(self):
         return "JOINT_VELOCITY"
 
+
 class LegacyMobileBaseJointVelocityController(MobileBaseJointVelocityController):
     """
     Legacy version of MobileBaseJointVelocityController, created to address
-    the recent change in the axis of the forward joint in the mobile base xml. 
+    the recent change in the axis of the forward joint in the mobile base xml.
     This controller is identical to the original MobileBaseJointVelocityController,
-    except that it dynamically checks the axis of the forward joint and reorders 
-    the input action accordingly if the forward axis is the y axis instead of the x axis. 
-    This allows for backwards compatibility with previously collected datasets 
+    except that it dynamically checks the axis of the forward joint and reorders
+    the input action accordingly if the forward axis is the y axis instead of the x axis.
+    This allows for backwards compatibility with previously collected datasets
     that were generated using older versions of the mobile base xml.
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-    
+
     def _check_forward_joint_reversed(self):
         # Detect the axis for the forward joint and dynamically reorder action accordingly.
         # This is needed because previous versions of the mobile base xml had different forward


robosuite/wrappers/data_collection_wrapper.py
```diff
@@ -69,10 +69,13 @@ def _start_new_episode(self):
...
-        # NOTE: was previously self.env.model.get_xml(). Was causing the following issue in rare cases:
+        # NOTE: was originally set to self.env.model.get_xml()
+        # That was causing the following issue in rare cases:
         # ValueError: Error: eigenvalues of mesh inertia violate A + B >= C
-        # switching to self.env.sim.model.get_xml() does not create this issue
-        self._current_task_instance_xml = self.env.sim.model.get_xml()
+        # then, switched to self.env.sim.model.get_xml() which does not create this issue
+        # however, that leads to subtle changes in the physics, such as fixture doors being harder to close
+        # so, reverting back to self.env.model.get_xml()
+        self._current_task_instance_xml = self.env.model.get_xml()
```
snasiriany: this may be a potentially big change and affect existing users. @Abhiram824 can you make it a flag and use `self.env.sim.model.get_xml()` by default? we should toggle the flag to use `self.env.model.get_xml()` only for robocasa for now. others can follow suit it they'd like

robosuite/controllers/parts/mobile_base/joint_vel.py
```diff
@@ -206,14 +206,29 @@ def set_goal(self, action, set_qpos=None):
...
-        # do the reverse of theta rotation
-        action[0] = x * np.cos(theta) + y * np.sin(theta)
-        action[1] = -x * np.sin(theta) + y * np.cos(theta)
+        forward_jnt = None
+        forward_jnt_axis = None
+        for jnt in self.joint_names:
+            if "joint_mobile_forward" in jnt:
+                forward_jnt = jnt
+                forward_jnt_axis = self.sim.model.jnt_axis[self.sim.model.joint_name2id(jnt)]
+                break
```
snasiriany: @Abhiram824 can you give some context for what's happening in this code block?

Good to also write some comments in the code to give other users context.

Did `forward_jnt_axis` change at some point?

snasiriany (commented):
leaving clarification questions and comments

robosuite/wrappers/visualization_wrapper.py
```diff
@@ -75,9 +75,10 @@ def __init__(self, env, indicator_configs=None):
 
         # Conduct a (hard) reset to make sure visualization changes propagate
         reset_mode = self.env.hard_reset
-        self.env.hard_reset = True
-        self.reset()
-        self.env.hard_reset = reset_mode
+        if self.env.hard_reset is not True:
```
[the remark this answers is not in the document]
Abhiram824: Im not 100% sure (I think you might've made this change), but it looks like logic is that we dont need the extra reset in the VisualizationWrapper because we already call reset right before we start collecting demonstrations.

I personally think we should undo this change though

robosuite/controllers/parts/mobile_base/joint_vel.py
snasiriany: It seems like a bit of a hack, because it's explicitly looking for a joined named `joint_mobile_forward`. Since this is pretty specific code, I would create a separate function for this, just for the purposes of "post-processing" the action to be `action = np.copy([action[i] for i in [1, 0, 2]])` instead. This will be cleaner!

Abhiram824: I agree that its very ugly, but for the time being i dont think there is much we can do about it. I will move it to a different function though

Abhiram824: updated

PR #807, opened 2026-05-15
fix unclosed file handle in get_xml

### LINKED ISSUE #801: get_xml() -> ResourceWarning: Enable tracemalloc to get the object allocation traceback
### System Info

```Shell
Robosuite version, Ubuntu 24.04
```

### Information

I am receiving this warning from `get_xml()` in `robosuite/utils/bindin_utils.py` whenever `get_state()` is called
```
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/bho36/Documents/code/FILM/./robosuite/robosuite/utils/binding_utils.py:506: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmpt8nu028r/model.xml' mode='r' encoding='UTF-8'>
  return open(filename).read()
```

My solution was to modify the code from this
```
    def get_xml(self):
        with TemporaryDirectory() as td:
            filename = os.path.join(td, "model.xml")
            ret = mujoco.mj_saveLastXML(filename.encode(), self._model)
            return open(filename).read()
```
to this.
```
    def get_xml(self):
        with TemporaryDirectory() as td:
            filename = os.path.join(td, "model.xml")
            ret = mujoco.mj_saveLastXML(filename.encode(), self._model)
            with open(filename) as f:
                return f.read()
```

### Reproduction

N/A

### Expected behavior

_No response_

darshmenon:
Fixes #801

`get_xml()` in `binding_utils.py` called `open(filename).read()` without a context manager, leaving the file handle open and causing `ResourceWarning: unclosed file` every time `get_state()` is called.

Fixed by wrapping in a `with` block so the file is properly closed after reading.

### DIFF
--- a/robosuite/utils/binding_utils.py
+++ b/robosuite/utils/binding_utils.py
@@ -505,7 +505,8 @@ class MjModel(metaclass=_MjModelMeta):
         with TemporaryDirectory() as td:
             filename = os.path.join(td, "model.xml")
             ret = mujoco.mj_saveLastXML(filename.encode(), self._model)
-            return open(filename).read()
+            with open(filename) as f:
+                return f.read()
 
     def get_joint_qpos_addr(self, name):
         """


kevin-thankyou-lin (approved):
lgtm

PR #816, opened 2026-07-09
Fix Wipe table height drifting across resets

### LINKED ISSUE #815: Wipe: _load_model adds delta_height to the persistent table_offset cumulatively — table height ramps linearly across resets; table_height_std never re-samples per episode
In `wipe.py`, `delta_height` is sampled once and then added to the persistent `self.table_offset` array on EVERY `_load_model` (i.e. every hard reset):

```python
# __init__ (line ~238)
self.delta_height = None  # sample variation in height done in load_model

# _load_model (lines ~537-539)
if self.delta_height is None:
    self.delta_height = self.rng.normal(self.table_height, self.table_height_std)
self.table_offset[2] += self.delta_height
```

`_reset_internal` never resets `delta_height` to `None`, and `self.table_offset` is the same array object across the environment's lifetime (created once in `__init__`). Two consequences:

**1. Linear height ramp across resets.** After k resets, `table_offset[2] = base + k * delta`. Minimal reproduction through the public API (robosuite 1.5.2, no internals touched):

```python
import robosuite as suite
from robosuite.environments.manipulation.wipe import DEFAULT_WIPE_CONFIG

cfg = dict(DEFAULT_WIPE_CONFIG)
cfg["table_height_std"] = 0.02   # the documented per-episode randomization knob
env = suite.make("Wipe", robots="Panda", has_renderer=False,
                 has_offscreen_renderer=False, use_camera_obs=False, task_config=cfg)
print(env.table_offset[2], env.delta_height)
for k in range(5):
    env.reset()
    print(env.table_offset[2], env.delta_height)
```

Output:

```
0.9280  delta_height = +0.0280
0.9560  +0.0280   <- same delta re-applied
0.9840  +0.0280
1.0121  +0.0280
1.0401  +0.0280
1.0681  +0.0280
```

In long training runs (SAC/TQC, ~100k steps, hundreds of resets) the ramp eventually produces a physically invalid table model (XML compile `ValueError`), killing the run — and once the ramp crosses into the invalid region, every subsequent reset fails as well.

**2. `table_height_std` does not deliver per-episode variation.** The config comment says "Standard deviation to sample different heigths of the table each episode", but the delta is sampled exactly once per environment instance — from episode 2 onward every episode sees the same (cumulatively re-applied) offset. The documented behavior never executes.

**Expected behavior:** each reset samples a fresh delta around the pristine base, e.g. keep `base_z` immutable and compute `table_offset[2] = base_z + self.rng.normal(self.table_height, self.table_height_std)` per `_load_model`.

**Workaround we currently use:** restore the pristine base and force a fresh sample before every reset:

```python
env.table_offset = base_offset.copy()
env.delta_height = None
```

**Environment:** robosuite 1.5.2 (PyPI wheel, byte-identical file), mujoco 3.x, Linux. The relevant lines are unchanged on current `master`. Happy to submit a PR with the fix if useful.

HaozheZhang6:
## What this does
Fixes #815 (🐛 Bug).

In `Wipe`, `self.table_offset[2] += self.delta_height` ran on every `_load_model` while `delta_height` was sampled only once (guarded by `is None`). Since `table_offset` is a persistent array, the table height accumulated a fixed delta each reset (a linear ramp), and `table_height_std` produced no per-episode variation after the first episode.

Now the base offset is stored once at init and each reset sets `table_offset[2] = base + N(table_height, table_height_std)`, so the height is re-sampled per episode and never drifts.

## How it was tested
Added `test_wipe_table_height_no_drift` in `tests/test_environments/test_wipe_table_height.py`: with `table_height_std=0.02` the resets no longer ramp, stay centered on the base height, and vary per episode; with the default `std=0.0` the height stays fixed (no regression).

Before: `table_offset[2]` over 6 resets ramped `0.930 -> 0.945 -> ... -> 1.004`. After: bounded and varying, e.g. `0.926, 0.923, 0.867, 0.912, ...`.

## How to checkout & try?
```bash
pytest -sx tests/test_environments/test_wipe_table_height.py
```


Abhiram824:
Hi, change looks good. can you remove the extra test file and I will approve

HaozheZhang6:
Removed, thanks!

Abhiram824 (approved):

